Issue #4, opened 2023-12-15
Fits don's include central value


lonbar:
One thing I noticed is that the `x_values` used to compute the function values don't include the central value (see [derivative_calculator.py at lines 55–66](https://github.com/nikosarcevic/Derivative-Calculator/blob/b332b9b61083e38a516a25a9aea58a8b128c53fa/derivative_calculator.py#L55-L66)). This isn't difficult to fix and may in fact not be necessary, but it is good to be aware of this.

nikosarcevic:
> One thing I noticed is that the `x_values` used to compute the function values don't include the central value (see [derivative_calculator.py at lines 55–66](https://github.com/nikosarcevic/Derivative-Calculator/blob/b332b9b61083e38a516a25a9aea58a8b128c53fa/derivative_calculator.py#L55-L66)). This isn't difficult to fix and may in fact not be necessary, but it is good to be aware of this.

yeah I noticed it, probs because I was adding the gaussian noise to it. can fix it when I do the upgrade with the broadcasting etc

PR #6, opened 2025-04-13
add 0% offset to percentage values

lonbar:
Solves #4.

### DIFF
--- a/derivative_calculator.py
+++ b/derivative_calculator.py
@@ -49,11 +49,19 @@ class DerivativeCalculator:
         tuple: the stem derivative.
         """
 
-        # The percentage values to use for the stem method
+        # The percentage offset values to use for the stem method
         # Note that this is an arbitrary choice and the values
         # can be changed as needed
-        percentages = [0.00625, 0.0125, 0.01875, 0.025, 0.0375, 0.05,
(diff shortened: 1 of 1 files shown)


nikosarcevic:
this is not gonna work out, im redoing it. it is not gonna be stable at zero in some functions so im adding some smart ways to go about it (fallback to stencil etc)

nikosarcevic:
After further development and testing, I’ve decided to close this PR. The original idea of adding 0.0 to the list of percentage offsets in the stem_method was meant to improve derivative accuracy at the center. However, this caused instability for functions that are not smooth or well-defined at $x=0$, such as:
- $\log(1 + x)$
- $\sqrt {x + 1}$
- Piecewise and ReLU-type functions

I've implemented a more robust solution in the [advanced-test-suite](https://github.com/nikosarcevic/Derivative-Calculator/pull/new/advanced-test-suite) branch, which includes:

- Smoothness checks to exclude 0% offset when problematic
- Adaptive fallback to the stencil method
- Extensive test coverage for over 25 functions
- Automated plotting of derivatives, tangents, and residuals
- Logging for traceability and debugging

The new version makes sense to me and behaves as expected across all tested cases. That said, I’d really appreciate your review to ensure I’m handling everything correctly and not overlooking edge cases.

Thanks!

PR #8, opened 2025-04-14
Add and Clean Up Derivative Testing Suite

nikosarcevic:
This pull request introduces a cleaned-up and comprehensive derivative testing suite. It covers both the Stem and Five-Point Stencil methods and evaluates their performance across a wide range of functions.

**Summary of Changes**

- Cleaned and modularized the testing code.
- Added full support for testing first and second derivatives.
- Tested against a broad set of analytic, piecewise, and non-smooth functions.
- Computed and logged absolute and relative differences between methods.
- Applied consistent tolerance-based pass/fail checks for result comparison.
- Marked clear discrepancies for functions with undefined behavior or known edge cases.

**Review Notes**

This PR is submitted separately from earlier development work to isolate the test suite. It is intended as a baseline validation framework. Please review:

- The structure and readability of the output
- Whether the tolerances and flagging thresholds are appropriate
- If additional functions or edge cases should be included
- Let me know if anything should be adjusted before merging.


lonbar:
I have done a broad-strokes pass and have the following initial feedback.

# The structure and readability of the output
I'm not sure I understand. The output of the library itself is the derivative. That's fine. Perhaps we could allow numpy arrays or something similar.

# The tolerances and flagging thresholds
I feel it is difficult to say much about this; as stated I think this depends on the requirements of the user.

# Inclusion of additional functions or edge cases
Again, I think it is difficult to say much at this point because the number of functions that can be differentiated is uncountably infinite. Ideally we only check whether the classes are properly initialised and the data correctly passed to external libraries, as in the end I expect the library will rely on others (scipy, numpy, etc) to do the actual computations.

I think that what's included in the PR is a good starting point. We can always extend. 

# Adjustments before merging

Some things stand out to me which could improve the code, moving forward:
* Make the code PEP-8 compliant (e.g. using `black`)
* Most files in this PR do not need to be included in the repository:
   * development and runtime directories such as `.idea` and `__pycache__`
   * the log files generated by the library
   * the directories containing images and test results generated at runtime
   * files and directories generated by the notebooks
   * the XML files
* The notebooks don't need to be included either. Their content could be included in a readme or example page. The code snippets can then be easily copy-pasted by users into their terminal/notebook/environment. If you do feel they are essential I would strongly suggest to first clear them of user-specific input (for example, log messages containing paths on your own laptop).
* It would be useful to include a `.gitignore` to help prevent such files being included in blanket commits
* The `HybridDerivativeCalculator` class mixes demonstration/testing logic with its core use (`calculate_derivatives_with_output_noise`). Wouldn't it make sense to split this off?

hybrid_derivative_calculator.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # Coefficients for derivatives
+        coeffs_dict = {
+            1: np.array([1, -8, 0, 8, -1]) / (12.0 * h),
+            2: np.array([-1, 16, -30, 16, -1]) / (12.0 * h ** 2),
+        }
+
+        coeffs = coeffs_dict.get(self.derivative_order)
+        if coeffs is None:
+            raise ValueError("Only derivative_order=1 or 2 is supported.")
```
lonbar, line 123: Doesn't this contradict the docstring, which states that third derivatives are supported?

hybrid_derivative_calculator.py
```diff
@@ -0,0 +1,185 @@
...
+            f0 = self.function(self.central_value)
+            f_minus = self.function(self.central_value - delta)
+            f_plus = self.function(self.central_value + delta)
+
+            if not np.isfinite([f0, f_minus, f_plus]).all():
+                return False
+
+            left = (f0 - f_minus) / delta
+            right = (f_plus - f0) / delta
+            return abs(left - right) < tol
```
lonbar, line 145: don't we want to use `np.abs`?

hybrid_derivative_calculator.py
```diff
@@ -0,0 +1,185 @@
...
+
+            try:
+                if method == 'stem':
+                    derivative = self.stem_method()
+                elif method == 'five_point_stencil':
+                    derivative = self.five_point_stencil_method()
+                else:
+                    raise ValueError("Invalid method. Choose 'stem' or 'five_point_stencil'.")
+            finally:
+                self.function = original_func  # always restore
```
lonbar, line 181: Don't we want to handle the case where the derivative method fails?

hybrid_derivative_calculator.py
```diff
@@ -0,0 +1,185 @@
...
+        # Generate timestamped log file name
+        timestamp = datetime.now().strftime("%Y%m%d")
+        if log_suffix:
+            self.log_file = os.path.join(log_dir, f"stem_debug_{timestamp}_{log_suffix}.log")
+        else:
+            self.log_file = os.path.join(log_dir, f"stem_debug_{timestamp}.log")
+
+    def stem_method(self, include_zero=True, fallback=True, debug=True):
+        """
+        Estimate the 1st, 2nd, or 3rd derivative using iterative polynomial fitting.
```
lonbar, line 49: The docstring allows for 3rd order derivative, but I don't see this in the subsequent code.

derivative_plotter.py
```diff
@@ -0,0 +1,197 @@
...
+
+            # Calculate derivatives
+            stem_derivatives.append(hdc_noisy.stem_method())
+            stencil_derivatives.append(hdc_noisy.five_point_stencil_method())
+
+        # Create two subplots for each method
+        fig, axes = plt.subplots(2, 1, figsize=(8, 6))
+
+        stem_color = 'yellowgreen'
+        stencil_color = 'hotpink'
```
lonbar, line 184: Couldn't we expose the plotting options to the user?

derivative_plotter.py
```diff
@@ -0,0 +1,197 @@
+import matplotlib.pyplot as plt
+import numdifftools as nd
+import numpy as np
+import os
+from hybrid_derivative_calculator import HybridDerivativeCalculator
```
lonbar, line 5: This doesn't comply with PEP-8's recommendation on [import statements](https://peps.python.org/pep-0008/#imports)

hybrid_derivative_calculator.py
```diff
@@ -0,0 +1,185 @@
+from datetime import datetime
+import numpy as np
+import os
+import warnings
```
lonbar, line 4: This doesn't comply with PEP-8's recommendation on import statements.

hybrid_derivative_calculator.py
nikosarcevic: Good point — I added a try/except around the derivative call so if it fails (e.g. no convergence or weird function behavior), it logs the issue and adds a np.nan instead. That way we don’t silently get garbage values, and we can handle or ignore the NaNs later.

Issue #18, opened 2025-08-08
Move derivative_order from __init__ to method argument in DerivativeKit


nikosarcevic:
Currently, DerivativeKit requires the derivative order at initialization, which limits flexibility. We propose moving `derivative_order` to the `.compute()` method so that the same object can be reused for computing derivatives of different orders.
 This simplifies integration with tools like ForecastKit and improves usability.

PR #23, opened 2025-08-08
Updated forecast_kit 

ctrendafilova:
The derivative order is now specified at the .compute() call rather than at DerivativeKit initialization.


lonbar:
These changes build on top of #18, right?

ctrendafilova:
@lonbar correct.

lonbar (commented):
Looks good to me! Merging this now would break things, as the main branch currently doesn't have these changes in `DerivativeKit`. @nikosarcevic do you want to merge #22 first?

PR #28, opened 2025-08-20
Remove __all__ imports

lonbar:
To my knowledge, __all__ is only used to allow users to import submodules from the package into the current namespace. This is considered bad practice, and is probably not something we want to support.

Further reading:
https://docs.python.org/3/tutorial/modules.html#importing-from-a-package


nikosarcevic (approved):
than kyou, please proceed.

PR #42, opened 2025-08-24
Update docstrings to Sphinx format

lonbar:
The module docstrings are written using various different conventions. As a result, the documentation pages aren't always properly rendered. I propose we rewrite the docstrings using the Sphinx docstring standard
(https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html).

I realise that other standards exist (see the linked page), and I have in fact enabled these styles in docs/conf.py using
 the Napoleon plugin
(https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html). I am not sure which is more desirable, and have mainly chosen the Sphinx convention in this commit because I was reading the Sphinx documentation. Feedback is much appreciated.

### DIFF
--- a/src/derivkit/adaptive_fit.py
+++ b/src/derivkit/adaptive_fit.py
@@ -13,35 +13,34 @@ class AdaptiveFitDerivative:
     """
     Computes derivatives using an adaptive polynomial fitting approach.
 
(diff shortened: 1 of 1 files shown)


lonbar:
Closing this as we don't want to use Sphinx-style docstrings. I'll reopen a different PR.

PR #56, opened 2025-09-21
Chore/repo structure

nikosarcevic:


.gitignore
```diff
@@ -12,3 +12,21 @@ dist
 notebooks/plots/*.png
 notebooks/plots/*.pdf
 docs/_build
+# --- notebooks ---
+.ipynb_checkpoints/
+notebooks/**/outputs/
+notebooks/**/.ipynb_checkpoints/
+notebooks/**/__pycache__/
```
lonbar, line 19: I would group these with the others,

.gitignore
```diff
@@ -12,3 +12,21 @@ dist
 notebooks/plots/*.png
 notebooks/plots/*.pdf
 docs/_build
+# --- notebooks ---
+.ipynb_checkpoints/
```
lonbar, line 14: I think you can just write something like
```
**.ipynb_checkpoints/
```
or something similar for all these notebook checkpoints

lonbar: That way you don't need the second entry (`notebooks/**/`etc)

.transfer_check
lonbar, line 1: What's the point of this file?

README.md
lonbar, line 1: This doesn't appear to go to working notebooks. I would leave it out for now.

binder/environment.yml
lonbar, line 1: Is there a reason to not put this under `.github/workflows`?

binder/environment.yml
```diff
@@ -0,0 +1,12 @@
+
+name: derivkit-binder
+channels: [conda-forge]
```
lonbar, line 3: I don't think we need conda, right? We could just use pip.

binder/environment.yml
```diff
@@ -0,0 +1,12 @@
...
+dependencies:
+  - python=3.11
+  - numpy
+  - scipy
+  - matplotlib
+  - ipykernel
+  - pip
```
lonbar, line 10: Can't these be read in from pyproject.toml?

.transfer_check
nikosarcevic: i think this was a log file, can be removed actually (AFTER we merge tho)

binder/environment.yml
nikosarcevic: dont know acutally

README.md
nikosarcevic: you mean add to gitignore?

binder/environment.yml
nikosarcevic: agreed.

nikosarcevic: deleted line 3

Issue #58, opened 2025-09-21
rename methods and arg in compute derivs functions


nikosarcevic:
currently we use `compute() `as a method to compute derivatives. it is too vague so decision is made to rename it to `differentiate()`.
the argument in differentiate to set the derivative order is currently `derivative_order `and sohuld be changed to just `order`.

PR #59, opened 2025-09-21
API rename:  and param cleanups

nikosarcevic:
Refs #58. Renames API (, param names), updates callers, and adjusts docs/tests.


nikosarcevic:
> Refs #58. Renames API (, param names), updates callers, and adjusts docs/tests.

i have also run pytests, all passed.

nikosarcevic:
i suspect the documentation will fail due to these changes. but maybe not, maybe it will pick it up automatically.

src/derivkit/adaptive_fit.py
```diff
@@ -8,7 +8,7 @@
 
 >>> derivative = AdaptiveFitDerivative(function, 1).differentiate(order=1)
 
-derivative is the derivative of function_to_differerentiate at value 1.
+derivative is the derivative of function at value 1.
```
lonbar, line 11: This is the line where I meant that it would perhaps be useful to add that it is the first derivative

lonbar (approved):

PR #79, opened 2025-09-25
List tox as dependency with several environments

lonbar:
Tox supports environments beyond just testing. Various different commands can be concatenated in a given environment. Tox is also able to recognise conditional dependencies, and will install these in the relevant environment before executing the commands.

This is very useful as it allows users to run things like the testing suite, the linter, or the docs building entirely through tox. The list of available environments can be shown by executing `tox list`, so it is easy for a user to work out what is available. To run, say, environment `docs`, the user only has to run `tox -e docs`. Additionally, this simplefies any workflows that are run on Github.

For this reason I am moving tox to the list of package requirements. It will then be available by default if a users installs derivkit via pip. For this to make sense, I am adding several workflows that are currently in use in tox' configuration.


nikosarcevic (approved):
do we also want to try uv for envs?

PR #102, opened 2025-09-27
Uv workflows

### LINKED ISSUE #84: Try out uv for environments
Let’s experiment with using [uv](https://github.com/astral-sh/uv) to manage Python environments.
It promises faster installs, lockfile support, and simpler workflows.

lonbar:
This adds uv to the CI workflows. It somewhat simplifies the configuration file and gives an improvement in the runtimes. Resolves #84.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,18 +19,17 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


nikosarcevic (approved):
looks but but i am not the best w tihs. so yes -- yolo!

PR #108, opened 2025-09-27
Move get_partial_function to utils

nikosarcevic:
Refactor: relocate helper to utils.py for reuse. No functional changes.


src/derivkit/forecasting/expansions.py
```diff
@@ -228,36 +229,6 @@ def function_to_diff2(y):
...
-                use as fixed inputs for all parameters except the one being
-                varied.
-
-        Returns:
-            callable: A function of a single variable, suitable for use in
-                differentiation.
-        """
-
-        def partial_function(x):
-            params = deepcopy(fixed_values)
```
lonbar, line 255: Since `deepcopy` is imported from the`copy` library the import in line 14 should be removed.

lonbar (approved):

Issue #80, opened 2025-09-25
sort the tests of derivkit


nikosarcevic:
separate the tests for deirvkit from adaptive and finite. each needs to be tested separately

PR #113, opened 2025-09-27
Tests(#80): add coverage for DerivativeKit and finite-difference

nikosarcevic:
This PR adds focused tests for the core derivative calculators in DerivKit, building on the recent refactor in #112.

**DerivativeKit**

- Tests constructor wiring with AdaptiveFitDerivative and FiniteDifferenceDerivative.
- Verifies get_used_points returns expected structures using monkeypatched diagnostics.

**FiniteDifferenceDerivative**

- Adds functional tests for analytic cases (sin, quadratic).
- Covers vector-valued outputs and invalid order handling.
- These tests strengthen core coverage and directly address issue #80.

**Notes:**

- AdaptiveFitDerivative is already exercised by the adaptive test suite; no duplication here.
- This PR should be merged after #112.

### DIFF
--- a/tests/test_derivative_kit.py
+++ b/tests/test_derivative_kit.py
@@ -1,0 +1,108 @@
+"""Tests for DerivativeKit wrapper."""
+
+import numpy as np
+
+from derivkit.derivative_kit import DerivativeKit
+
+
+def test_constructor_wires_adaptive_and_finite(monkeypatch):
+    """Both calculators are constructed with the same (function, x0)."""
+    calls = {"adaptive": None, "finite": None}
+
+    class FakeAdaptive:
+        def __init__(self, function, x0, **kwargs):  # noqa: D401
+            calls["adaptive"] = (function, x0, kwargs)
+
+        def differentiate(self, **kwargs):  # not used here
+            return 0.0
+
+    class FakeFinite:
+        def __init__(self, function, x0, **kwargs):  # noqa: D401
+            calls["finite"] = (function, x0, kwargs)
+
+    # Patch module-local symbols used by DerivativeKit
+    monkeypatch.setattr(
+        "derivkit.derivative_kit.AdaptiveFitDerivative",
+        FakeAdaptive,
+        raising=True,
+    )
+    monkeypatch.setattr(
+        "derivkit.derivative_kit.FiniteDifferenceDerivative",
+        FakeFinite,
+        raising=True,
+    )
+
+    def f(x):
+        return x
+
+    dk = DerivativeKit(f, 0.5)
+
+    assert calls["adaptive"][0] is f and np.isclose(calls["adaptive"][1], 0.5)
+    assert calls["finite"][0] is f and np.isclose(calls["finite"][1], 0.5)
+    assert isinstance(dk.adaptive, FakeAdaptive)
+    assert isinstance(dk.finite, FakeFinite)
+
+
+def test_get_used_points_plumbs_diagnostics(monkeypatch):
+    """get_used_points must request diagnostics and unpack them correctly."""
+    captured = {"called": False, "kwargs": None}
+
+    class FakeAdaptive:
+        def __init__(self, *args, **kwargs):  # noqa: D401
+            pass
+
+        def differentiate(self, *, order, diagnostics, n_workers, **_):
+            captured["called"] = True
+            captured["kwargs"] = {
+                "order": order,
+                "diagnostics": diagnostics,
+                "n_workers": n_workers,
+            }
+
+            # Diagnostics payload matching DerivativeKit expectations
+            M, C = 5, 1
+            diag = {
+                "x_all": np.linspace(0.0, 1.0, M),
+                "y_all": np.linspace(10.0, 20.0, M).reshape(M, C),
+                "x_used": [np.array([0.0, 0.25, 0.5])],
+                "y_used": [np.array([10.0, 12.5, 15.0])],
+                "used_mask": [np.array([True, False, True, True, False])],
+            }
+            return 0.0, diag
+
+    class FakeFinite:
+        def __init__(self, *args, **kwargs):  # noqa: D401
+            pass
+
+    monkeypatch.setattr(
+        "derivkit.derivative_kit.AdaptiveFitDerivative",
+        FakeAdaptive,
+        raising=True,
+    )
+    monkeypatch.setattr(
+        "derivkit.derivative_kit.FiniteDifferenceDerivative",
+        FakeFinite,
+        raising=True,
+    )
+
+    dk = DerivativeKit(lambda x: x, 0.0)
+    x_all, y_all, x_used, y_used, used_mask = dk.get_used_points(
+        order=2, n_workers=3
+    )
+
+    # Ensure diagnostics=True and kwargs forwarded
+    assert captured["called"] is True
+    assert captured["kwargs"] == {
+        "order": 2,
+        "diagnostics": True,
+        "n_workers": 3,
+    }
+
+    # Shapes & basic value checks
+    assert x_all.shape == (5,)
+    assert y_all.shape == (5,)
+    assert x_used.shape == (3,)
+    assert y_used.shape == (3,)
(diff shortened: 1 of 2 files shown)


lonbar:
This branch probably has to be rebased onto main after merging #112, as it shares a lot of its changes.

tests/test_derivative_kit.py
```diff
@@ -0,0 +1,108 @@
+"""Tests for DerivativeKit wrapper."""
+
+import numpy as np
+
+from derivkit.derivative_kit import DerivativeKit
+
+
+def test_constructor_wires_adaptive_and_finite(monkeypatch):
+    """Both calculators are constructed with the same (function, x0)."""
```
lonbar, line 9: This doesn't describe the test.

tests/test_derivative_kit.py
```diff
@@ -0,0 +1,108 @@
...
+    dk = DerivativeKit(f, 0.5)
+
+    assert calls["adaptive"][0] is f and np.isclose(calls["adaptive"][1], 0.5)
+    assert calls["finite"][0] is f and np.isclose(calls["finite"][1], 0.5)
+    assert isinstance(dk.adaptive, FakeAdaptive)
+    assert isinstance(dk.finite, FakeFinite)
+
+
+def test_get_used_points_plumbs_diagnostics(monkeypatch):
+    """get_used_points must request diagnostics and unpack them correctly."""
```
lonbar, line 47: This doesn't describe the test

tests/test_derivative_kit.py
```diff
@@ -0,0 +1,108 @@
...
+    captured = {"called": False, "kwargs": None}
+
+    class FakeAdaptive:
+        def __init__(self, *args, **kwargs):  # noqa: D401
```
lonbar, line 51: I don't think this is the right preprocessing macro, and I think you need D107 instead.

lonbar: Alternatively, since the other modules don't appear to have docstrings, perhaps you don't need this at all.

tests/test_finite_difference.py
```diff
@@ -18,3 +18,42 @@ def test_invalid_order_finite():
...
         DerivativeKit(lambda x: x, 1.0).finite.differentiate(order=5)
+
+
+def test_fd_second_derivative_quadratic_constant():
+    """Second derivative of a quadratic is constant: d²/dx² (ax²+bx+c) = 2a."""
+    a, b, c = 3.0, -1.0, 2.0
+
+    def myfunc(x):
+        return a * x**2 + b * x + c
+
```
lonbar, line 29: is it necessary to nest this function?

tests/test_finite_difference.py
```diff
@@ -18,3 +18,42 @@ def test_invalid_order_finite():
...
+    est = DerivativeKit(myfunc, x0=0.3).finite.differentiate(order=2, num_points=5)
+    assert np.isclose(est, 2 * a, rtol=1e-3, atol=1e-8)
+
+
+def test_vector_output_returns_1d_array():
+    """Multi-component output returns 1D NumPy array of derivatives."""
+
+    def myfunc(x):
+        return np.array([x**2, 2 * x])
```
lonbar, line 38: Is it necessary to nest this function?

tests/test_derivative_kit.py
```diff
@@ -0,0 +1,108 @@
...
+                "x_all": np.linspace(0.0, 1.0, M),
+                "y_all": np.linspace(10.0, 20.0, M).reshape(M, C),
+                "x_used": [np.array([0.0, 0.25, 0.5])],
+                "y_used": [np.array([10.0, 12.5, 15.0])],
+                "used_mask": [np.array([True, False, True, True, False])],
+            }
+            return 0.0, diag
+
+    class FakeFinite:
+        def __init__(self, *args, **kwargs):  # noqa: D401
```
lonbar, line 74: Same comment as in the `__init__` module of `FakeAdaptive`.

tests/test_derivative_kit.py
```diff
@@ -0,0 +1,108 @@
...
+            pass
+
+    monkeypatch.setattr(
+        "derivkit.derivative_kit.AdaptiveFitDerivative",
+        FakeAdaptive,
+        raising=True,
+    )
```
lonbar, line 81: Perhaps it is more appropriate to place this directly after the definition of `FakeAdaptive`?

tests/test_derivative_kit.py
```diff
@@ -0,0 +1,108 @@
...
+        order=2, n_workers=3
+    )
+
+    # Ensure diagnostics=True and kwargs forwarded
+    assert captured["called"] is True
+    assert captured["kwargs"] == {
+        "order": 2,
+        "diagnostics": True,
+        "n_workers": 3,
+    }
```
lonbar, line 99: Perhaps it makes sense to define variables for `order` and `n_workers`, but both uses are close enough together that it may not be essential.

tests/test_derivative_kit.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # Shapes & basic value checks
+    assert x_all.shape == (5,)
+    assert y_all.shape == (5,)
+    assert x_used.shape == (3,)
+    assert y_used.shape == (3,)
+    assert used_mask.shape == (5,)
```
lonbar, line 106: These depend on hard-coded values in `FakeAdaptive`, so it may make sense to define a variable somewhere.

tests/test_derivative_kit.py
```diff
@@ -0,0 +1,108 @@
...
+            diag = {
+                "x_all": np.linspace(0.0, 1.0, M),
+                "y_all": np.linspace(10.0, 20.0, M).reshape(M, C),
+                "x_used": [np.array([0.0, 0.25, 0.5])],
+                "y_used": [np.array([10.0, 12.5, 15.0])],
+                "used_mask": [np.array([True, False, True, True, False])],
+            }
+            return 0.0, diag
+
+    class FakeFinite:
```
lonbar, line 73: Is this class used anywhere in this test?

tests/test_derivative_kit.py
nikosarcevic: pushed.

tests/test_finite_difference.py
nikosarcevic: pusged.

tests/test_derivative_kit.py
nikosarcevic: removed and pushed.

tests/test_derivative_kit.py
nikosarcevic: yeah makes sense, I’ve moved the monkeypatch right under the fake class defs so it’s clearer. Keeps the scope tight and easier to read.
i also changed varsity M and C to m and c lowercase. now pushed.

tests/test_derivative_kit.py
nikosarcevic: done and pushed.

tests/test_derivative_kit.py
nikosarcevic: yeah good catch. now removed nad pushed (two pushed as i forgot to rmeove the monkeypatch)

tests/test_derivative_kit.py
lonbar: It still doesn't describe the test. It just says what `DerivativeKit` does.

tests/test_derivative_kit.py
lonbar: I don't think this is actually what you did, but I'm happy for it to stay as-is.

tests/test_derivative_kit.py
nikosarcevic: changed it, pushed

tests/test_derivative_kit.py
nikosarcevic: fixed and pushed.

tests/test_finite_difference.py
nikosarcevic: done and pushed.

lonbar (approved):

PR #127, opened 2025-09-30
fisher bias implementation 

nikosarcevic:
- Implements LikelihoodExpansion.fisher_bias:
- Accepts either delta or (datavec_with, datavec_without) and builds Δμ internally.
- Reuses existing Fisher builder; prefers public jacobian with internal fallback.
- Adds small helpers. No breaking changes.


src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+        )
+
+        # b = J^T C^{-1} Δμ with J^T = d1
+        bias_vec = d1 @ (inv_cov @ delta_mu)
+
+        try:
+            delta_theta = np.linalg.solve(fisher, bias_vec)
+        except np.linalg.LinAlgError:
+            warnings.warn("Fisher is singular; using pseudoinverse.", RuntimeWarning)
+            delta_theta = np.linalg.pinv(fisher, rcond=1e-12) @ bias_vec
```
lonbar, line 362: Since this has shown up once or twice before perhaps it makes sense to put this in a function in utils

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+
+        if d1.shape != (n_params, n_obs):
+            raise ValueError(f"d1 has shape {d1.shape}, expected {(n_params, n_obs)}.")
+        fisher = self._build_fisher(d1, inv_cov)  # (P, P)
+
+        delta_mu = self._build_delta_vector(
+            delta=delta, data_with=datavec_with, data_without=datavec_without, n_obs=n_obs
+        )
+
+        # b = J^T C^{-1} Δμ with J^T = d1
```
lonbar, line 355: Don't use fancy unicode shenanigans. Use full sentences as outlined on the [Contributing](https://www.derivkit.org/contributing.html#contributing) page.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+                (corr, ell). Used only if `delta` is None.
+            datavec_without: Data vector without the systematic. Same shape rules
+                as `data_with`. Used only if `delta` is None.
+            n_workers: Number of workers passed to derivative routines.
+
+        Returns:
+            dict: A dictionary with:
+                - "F": Fisher matrix, shape (P, P).
+                - "b": Bias “force” vector Jᵀ C⁻¹ Δμ, shape (P,).
+                - "delta_theta": Parameter bias estimate Δθ, shape (P,).
```
lonbar, line 325: * Types are missing.
* Fisher bias does not rely on the exact nature of the underlying theory so I feel it is a bit strange to mention that here. Can't it be implemented in an agnostic way (using contractions of numpy arrays or whatever)?
* The descriptions of these entries assume context that is not present elsewhere. That should be explicitly stated either here or after the summary line above.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+                or if neither `delta` nor the pair (`data_with`, `data_without`)
+                is provided.
+        """
+        n_obs, n_params = self.n_observables, self.n_parameters
+
+        inv_cov = self._inv_cov()  # (N, N) shape
+        if inv_cov.shape != (n_obs, n_obs):
+            raise ValueError(f"inv_cov has shape {inv_cov.shape}, expected {(n_obs, n_obs)}")
+
+        # d1: (P,N). Prefer public Jacobian (N,P) -> transpose; fallback to internal.
```
lonbar, line 338: Use descriptive comments in plain language, as outlined on the [Contributing](https://www.derivkit.org/contributing.html#contributing) page.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+        try:
+            jac = jacobian(self.function, self.theta0, n_workers=n_workers)  # (N, P) shape
+            if jac.shape != (n_obs, n_params):
+                raise ValueError(f"Jacobian has shape {jac.shape}, expected {(n_obs, n_params)}.")
```
lonbar, line 342: Given that the shape of the result is checked again later this feels redundant.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+            d1 = jac.T  # (P, N) shape here
+        except Exception:
+            d1 = self._get_derivatives(order=1, n_workers=n_workers)  # (P, N) shape niko sanity check
+
+        if d1.shape != (n_params, n_obs):
+            raise ValueError(f"d1 has shape {d1.shape}, expected {(n_params, n_obs)}.")
+        fisher = self._build_fisher(d1, inv_cov)  # (P, P)
```
lonbar, line 349: I feel it would make more sense to have a pre-computed fisher matrix, as it is possible to do forecasting and fisher bias separately. You'd be doing double to work if you'd have to recompute the fisher matrix each time.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+
+        delta_mu = self._build_delta_vector(
```
lonbar, line 351: Please use descriptive variable names.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+            raise ValueError(f"inv_cov has shape {inv_cov.shape}, expected {(n_obs, n_obs)}")
+
+        # d1: (P,N). Prefer public Jacobian (N,P) -> transpose; fallback to internal.
+        try:
+            jac = jacobian(self.function, self.theta0, n_workers=n_workers)  # (N, P) shape
+            if jac.shape != (n_obs, n_params):
+                raise ValueError(f"Jacobian has shape {jac.shape}, expected {(n_obs, n_params)}.")
+            d1 = jac.T  # (P, N) shape here
+        except Exception:
+            d1 = self._get_derivatives(order=1, n_workers=n_workers)  # (P, N) shape niko sanity check
```
lonbar, line 345: Please use descriptive variable names.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+        # b = J^T C^{-1} Δμ with J^T = d1
+        bias_vec = d1 @ (inv_cov @ delta_mu)
+
+        try:
+            delta_theta = np.linalg.solve(fisher, bias_vec)
+        except np.linalg.LinAlgError:
+            warnings.warn("Fisher is singular; using pseudoinverse.", RuntimeWarning)
+            delta_theta = np.linalg.pinv(fisher, rcond=1e-12) @ bias_vec
+
+        return {"F": fisher, "b": bias_vec, "delta_theta": delta_theta}
```
lonbar, line 364: As I mentioned before, it feels redundant to compute the fisher matrix here.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+            delta: Precomputed mismatch Δμ. If provided, it is validated and used
+                directly (1D length N or 2D to be flattened).
+            data_with: Data vector with the systematic present. Used only if
+                `delta` is None. 1D length N or 2D to be flattened.
+            data_without: Data vector without the systematic. Same shape rules as
+                `data_with`. Used only if `delta` is None.
+            n_obs: Expected length N of the flattened data vector.
+
+        Returns:
+            np.ndarray: Δμ as a 1D vector of length N.
```
lonbar, line 413: Please give descriptions in full sentences without using unicode shenanigans.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+            delta_theta = np.linalg.pinv(fisher, rcond=1e-12) @ bias_vec
+
+        return {"F": fisher, "b": bias_vec, "delta_theta": delta_theta}
+
+    def _flatten_data_vector(self, data: np.ndarray, n_obs: int) -> np.ndarray:
+        """Return a 1D data vector of length N from 1D/2D input.
+
+        Args:
+            data: Input data. Either 1D of length N or 2D (corr, ell), which will be
+                flattened in C-order.
```
lonbar, line 371: As above, it feels inappropriate to refer to a specific use-case here

lonbar (changes requested):

lonbar:
What I forgot to add to this is that tests are missing.

src/derivkit/forecasting/expansions.py
nikosarcevic: removed

src/derivkit/forecasting/expansions.py
nikosarcevic: ok so hmm i have tried to fix the typing and some varaible naming cleanup in fdae359c0291860cbcc0e438db9b9277d13a8307.

but the problem is not sure how to approach other comments, i tihnk this is best if we sit down together and solve.

src/derivkit/forecasting/expansions.py
nikosarcevic: see my prevous response: we should sit down together to do this i think

src/derivkit/forecasting/expansions.py
nikosarcevic: would this delta_vector work? pushed in 9130816c0e117fef6e296585912e1b7e39f71f86

src/derivkit/forecasting/expansions.py
nikosarcevic: yeah we can figure it out

nikosarcevic:
> What I forgot to add to this is that tests are missing.

yes will add them in another pr. need to create an issue

src/derivkit/forecasting/expansions.py
lonbar: That would require you to explain what the delta is about. But since it's built using the `_build_delta_vector` method I think it's fine.

src/derivkit/forecasting/expansions.py
```diff
@@ -304,6 +305,120 @@ def _build_dali(self, d1, d2, invcov):
...
+            n_obs: Expected length N of the flattened data vector.
+
+        Returns:
+            np.ndarray: Flattened 1D array of length N.
+
+        Raises:
+            ValueError: If `x` is not 1D or 2D, or if the flattened size does not
```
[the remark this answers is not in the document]
lonbar: Added a comment.

src/derivkit/forecasting/expansions.py
lonbar: That's a good addition, thanks. I still feel that the `(corr, ell)` is out of the blue that seems out of place here.

nikosarcevic:
> What I forgot to add to this is that tests are missing.

hmm yeah should i do it now in this same branch or you want it separately

ctrendafilova:
There has already been thorough discussion on this, so I don't have much to add.

Regarding the redundancy of computing the Fisher matrix again (if it has already been computed) -- I just want to point out that the previous step of computing the observable derivatives is also redundant, since that piece is utilized directly in computing the bias vector.

I guess one option is to add yet another set of input options (or maybe a separate method entirely) that can build the Fisher bias if you pass a pre-computed Fisher matrix and derivative matrix. I don't know if that is ideal.

src/derivkit/forecasting/expansions.py
```diff
@@ -316,6 +317,144 @@ def _build_dali(self, d1, d2, invcov):
...
+        if self.cov.shape != (n_obs, n_obs):
+            raise ValueError(
+                f"covariance shape {self.cov.shape} must be (n, n) = {(n_obs, n_obs)} from the Jacobian."
+            )
+        if fisher_matrix.shape != (n_params, n_params):
+            raise ValueError(
+                f"fisher_matrix shape {fisher_matrix.shape} must be (p, p) = {(n_params, n_params)} from the Jacobian."
+            )
+
+        # delta_nu → 1D length n
```
lonbar: If this describes the flattening mentioned in the docs, please make this explicit. Comments should be written in grammatical English using full sentences.

src/derivkit/forecasting/expansions.py
```diff
@@ -316,6 +317,144 @@ def _build_dali(self, d1, d2, invcov):
...
+        """
+        fisher_matrix = np.asarray(fisher_matrix, dtype=float)
+        if fisher_matrix.ndim != 2 or fisher_matrix.shape[0] != fisher_matrix.shape[1]:
+            raise ValueError(f"fisher_matrix must be square; got shape {fisher_matrix.shape}.")
+
+        # Jacobian has shape (n, p)
+        j_matrix = jacobian(self.function, self.theta0, n_workers=n_workers)
+        n_obs, n_params = j_matrix.shape
+
+        # Check C and F shapes against J
```
lonbar:
```suggestion
        # Check shapes of the covariance and Fisher matrices against the Jacobian
```

src/derivkit/forecasting/expansions.py
```diff
@@ -316,6 +317,144 @@ def _build_dali(self, d1, d2, invcov):
...
+            delta_nu,
+            rcond=rcond,
+            assume_symmetric=True,
+            warn_context="covariance solve",
+        )
+
+        # Compute the bias vector
+        bias_vec = j_matrix.T @ cinv_delta
+
+        # Compute the parameter shifts
```
lonbar: This comment adds no extra information and should be removed[](url)

src/derivkit/forecasting/expansions.py
```diff
@@ -316,6 +317,144 @@ def _build_dali(self, d1, d2, invcov):
...
+            *,
+            dtype: type = float,
+    ) -> np.ndarray:
+        """Compute the difference between two data vectors for use in Fisher-bias estimates.
+
+        This function takes two data vectors—typically one with a systematic included
+        and one without—and returns their difference as a 1D array that matches the
+        expected number of observables in this instance. It works with both 1D inputs
+        and 2D arrays (for example, correlation × ell) and flattens 2D arrays using the
+        chosen memory order.
```
lonbar: It isn't entirely clear to me what is meant with "chosen memory order". It would be good to clarify this.

src/derivkit/forecasting/expansions.py
```diff
@@ -316,6 +317,144 @@ def _build_dali(self, d1, d2, invcov):
...
+
+        We standardize on row-major (“C”) flattening of 2D arrays, where the last
```
lonbar:
```suggestion
        We standardize on row-major flattening of 2D arrays, where the last
```

src/derivkit/forecasting/expansions.py
```diff
@@ -316,6 +317,144 @@ def _build_dali(self, d1, d2, invcov):
...
+        axis varies fastest. Ensure any data vectors and associated covariances are
+        constructed with the same convention for consistent results.
```
lonbar: Since this is written in the imperative it seems like the user needs to take care of this. Is this true?

src/derivkit/utils.py
```diff
@@ -195,3 +197,52 @@ def partial_function(x):
...
 
     return partial_function
+
+
+def solve_or_pinv(matrix: np.ndarray, vector: np.ndarray, *, rcond: float = 1e-12,
+                  assume_symmetric: bool = True, warn_context: str = "linear solve") -> np.ndarray:
+    """Solve ``system_matrix @ x = rhs`` with pseudoinverse fallback.
+
+    If ``assume_symmetric`` is True (e.g., Fisher matrices), attempt a
+    Cholesky-based solve. On failure (not SPD / singular), warn and fall
```
lonbar: You may want to add that SPD stands for symmetric positive definite.

src/derivkit/forecast_kit.py
```diff
@@ -47,3 +47,29 @@ def dali(self, *, n_workers: int = 1):
...
+            delta_nu=delta_nu,
+            n_workers=n_workers,
+            rcond=rcond
+        )
+
+    def delta_nu(self,
+                 data_with: np.ndarray,
+                 data_without: np.ndarray,
+                 ):
+        """Return the delta_nu vector with shape (N,) with N being the number of observables."""
```
lonbar: You may want to add a return type hint. THe docstring does not explain the inputs and outputs.

src/derivkit/forecasting/expansions.py
nikosarcevic: it should be C there as it is a numpy thing. i do not want to remove that C.

lonbar: It means "as used in C, the programming language". Stating that it is row-major is sufficient, right?

src/derivkit/forecasting/expansions.py
nikosarcevic: ok i can do it -- done in 0b860ac6dafba96a2d259e08264862ec8901936f

src/derivkit/forecast_kit.py
```diff
@@ -47,3 +47,29 @@ def dali(self, *, n_workers: int = 1):
...
         return self._lx.get_forecast_tensors(forecast_order=2, n_workers=n_workers)
+
+    def fisher_bias(self,
+                    *,
+                    fisher_matrix: np.ndarray = None,
+                    delta_nu: np.ndarray = None,
+                    n_workers: int = 1,
+                    rcond: float = 1e-12
+                    ):
+        """Return the Fisher bias vector with shape (P,) with P being the number of model parameters."""
```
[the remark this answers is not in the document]
lonbar: While that particular commit doesn't add a return type hint, I can see that it is there now.

src/derivkit/forecasting/expansions.py
```diff
@@ -316,6 +317,144 @@ def _build_dali(self, d1, d2, invcov):
...
+        return bias_vec, delta_theta
+
+    def build_delta_nu(
+            self,
+            data_with: np.ndarray,
+            data_without: np.ndarray,
+            *,
+            dtype: type = float,
+    ) -> np.ndarray:
+        """Compute the difference between two data vectors for use in Fisher-bias estimates.
```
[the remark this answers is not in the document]
lonbar: Looking at that commit, I suggest replacing
```
Compute the difference between two data vectors. for use in Fisher-bias estimates.
```
with
```
Compute the difference between two data vectors.
```

src/derivkit/forecasting/expansions.py
lonbar: fair enough.

PR #129, opened 2025-09-30
Add Jacobian tests; fix jacobian shape handling and docstrings

nikosarcevic:
See commit body for details; includes tests + fixes as described.

### DIFF
--- a/src/derivkit/forecasting/calculus.py
+++ b/src/derivkit/forecasting/calculus.py
@@ -1,7 +1,8 @@
 """Differential calculus helpers."""
 
-import numpy as np
 from collections.abc import Callable
+
+import numpy as np
 
 from derivkit.derivative_kit import DerivativeKit
 from derivkit.utils import get_partial_function
@@ -87,16 +88,14 @@ def jacobian(function, theta0, n_workers=1):
         raise ValueError("theta0 must be a non-empty 1D array.")
 
     # n_workers controls inner 1D differentiation (not across parameters).
-    jacobian = np.array(
-        [_jacobian_component(function, theta0, i, n_workers) for i in range(theta0.size)],
-        dtype=float,
-    )
-    if not np.isfinite(jacobian).all():
+    cols = [_jacobian_component(function, theta0, i, n_workers) for i in range(theta0.size)]
+    jac = np.stack(cols, axis=1)  # (m_outputs, n_params)
+    if not np.isfinite(jac).all():
         raise FloatingPointError("Non-finite values encountered in jacobian.")
-    return jacobian.T
+    return jac
 
 def _jacobian_component(function: Callable, theta0: np.ndarray, i: int, n_workers: int) -> np.ndarray:
-    """Compute one column of the jacobian of a scalar-valued function.
+    """Compute one column of the jacobian of a vector-valued function.
 
     Helper used by ``jacobian``. Wraps ``function`` into a single-variable
     callable via ``derivkit.utils.get_partial_function`` and differentiates it
@@ -120,7 +119,7 @@ def _jacobian_component(function: Callable, theta0: np.ndarray, i: int, n_worker
     partial_vec = get_partial_function(function, i, theta0)
 
     kit = DerivativeKit(partial_vec, theta0[i])
-    return kit.adaptive.differentiate(order=1, n_workers=n_workers)
+    return np.atleast_1d(np.asarray(kit.adaptive.differentiate(order=1, n_workers=n_workers), dtype=float))
 
 def hessian(*args, **kwargs):
     """This is a placeholder for a Hessian computation function."""
--- a/tests/test_calculus.py
+++ b/tests/test_calculus.py
@@ -1,9 +1,178 @@
-"""Unit tests for forecasting/calculus.py."""
+"""Unit tests for forecasting/calculus.py (Jacobian)."""
 
+import numpy as np
 import pytest
 
+from derivkit.forecasting.calculus import jacobian
 
-@pytest.mark.skip(reason="Pending calculus module: tracked in #116")
-def test_calculus_todo():
-    """Placeholder until forecasting/calculus.py exists; tests will be added then."""
-    pass
+
+def test_jacobian_linear_map():
+    """Test jacobian on a linear map."""
+    # f(theta) = A @ theta  =>  J = A (constant)
+    A = np.array([[1.0, -2.0, 0.5],
+                  [0.0,  3.0, 1.0]])
+    def f(th):
+        th = np.asarray(th, dtype=float)
+        return A @ th
+    theta0 = np.array([0.3, -0.7, 1.2])
+    J = jacobian(f, theta0, n_workers=1)
+    assert J.shape == (A.shape[0], theta0.size)
+    assert np.allclose(J, A, atol=1e-12, rtol=0.0)
+
+
+def test_jacobian_analytic():
+    """Test jacobian on a function with known analytic Jacobian."""
+    # f([x,y]) = [x^2, sin(y), x*y]
+    # J = [[2x,   0 ],
+    #      [ 0 , cos y],
+    #      [ y ,  x  ]]
+    def f(th):
+        x, y = th
+        return np.array([x**2, np.sin(y), x*y], dtype=float)
+    x0, y0 = 0.4, -0.2
+    theta0 = np.array([x0, y0], dtype=float)
+    J = jacobian(f, theta0, n_workers=2)
+    J_true = np.array([[2*x0, 0.0],
+                       [0.0,  np.cos(y0)],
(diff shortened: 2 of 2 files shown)


src/derivkit/forecasting/calculus.py
```diff
@@ -73,33 +85,60 @@ def _grad_component(
...
 
 
-def jacobian(function, theta0, n_workers=1):
-    """Returns the jacobian of a vector-valued function.
+def jacobian(
+    function: Callable[[ArrayLike], ArrayLike | float],
+    theta0: ArrayLike,
+    n_workers: int | None = 1,
+) -> NDArray[np.floating]:
+    """Jacobian of a vector-valued function where each column is the derivative wrt one parameter.
```
lonbar:
```suggestion
    """Computes the Jacobian of a vector-valued function.
    
    Each column in the Jacobian is the derivative with respect to one parameter.
```

src/derivkit/forecasting/calculus.py
```diff
@@ -73,33 +85,60 @@ def _grad_component(
...
-        ],
-        dtype=float,
-    )
-    if not np.isfinite(jacobian).all():
-        raise FloatingPointError("Non-finite values encountered in jacobian.")
-    return jacobian.T
+    y0 = np.atleast_1d(function(theta0))
+    m, n = int(y0.size), int(theta0.size)
+
+    # Fail fast if the base model output is non-finite
```
lonbar:
```suggestion
```

src/derivkit/forecasting/calculus.py
```diff
@@ -281,3 +320,62 @@ def _check_scalar_valued(function, theta0: np.ndarray, i: int, n_workers: int):
...
+
+    Returns:
+        A 1D array representing the derivative of the function with respect to theta[idx].
+    """
+    theta0_x = deepcopy(np.atleast_1d(theta0))
+    f_i = get_partial_function(function, idx, theta0_x)   # this sets theta[idx]=y
+    kit = DerivativeKit(f_i, theta0_x[idx])
+    # Keep inner serial to avoid nested pools; adaptive can still batch-eval internally.
+    inner_workers = 1
+    gi = kit.adaptive.differentiate(order=1, n_workers=inner_workers)
```
lonbar:
```suggestion
    gi = kit.adaptive.differentiate(order=1)
```
Since the default behaviour of the `adaptive` method is to not use multiprocessing.

tests/test_calculus_jacobian.py
```diff
@@ -0,0 +1,190 @@
...
+    jacg = jacobian(g, theta0, n_workers=1)
+    u0 = mat @ theta0
+    jacf = jacobian(f_base2, u0, n_workers=1)
+    jacg_theory = jacf @ mat
+    assert np.allclose(jacg, jacg_theory, atol=8e-5, rtol=5e-6)
+
+
+def test_jacobian_raises_on_nonfinite_output():
+    """Test jacobian raises on non-finite output components."""
+    with pytest.raises((FloatingPointError, ValueError)):
```
lonbar: The docstring of `jacobian` states that a non-finite results raise a `FloatingPointError`. `ValueError`s are raised when the parameter list is empty.

tests/test_calculus_jacobian.py
```diff
@@ -0,0 +1,190 @@
...
+    jac_true = np.array([[2*theta0[0], 1.0]])
+    assert jac.shape == (1, 2)
+    assert np.allclose(jac, jac_true, atol=1e-6, rtol=1e-6)
+
+
+def test_jacobian_workers_invariance():
+    """Test that jacobian results are invariant to n_workers."""
+    theta0 = np.array([0.4, -0.2], dtype=float)
+    jac1 = jacobian(f_analytic_2d, theta0, n_workers=1)
+    jac2 = jacobian(f_analytic_2d, theta0, n_workers=3)
```
lonbar: This assumes that the test runner will always have 3 workers available. Is this guaranteed?

src/derivkit/forecasting/calculus.py
```diff
@@ -73,33 +85,60 @@ def _grad_component(
...
+    # Fail fast if the base model output is non-finite
+    if not np.isfinite(y0).all():
+        raise FloatingPointError("Non-finite values in model output at theta0.")
+
+    try:
+        work = int(n_workers) if n_workers is not None else 1
+    except (TypeError, ValueError):
+        work = 1
+    if work < 1:
+        work = 1
```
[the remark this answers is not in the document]
nikosarcevic: what you are saying is tihnk  sets work = 1 whenever n_workers >= 1 and that is the opposite of what we watn here

plus it would probs craash on none.

i have this now 770563786d1a9e73a71efcdb8549912d405e03b1

(this also fixes import issue in expansions whihc called imort jacobian instead of import build_jacobian)

src/derivkit/forecasting/calculus.py
```diff
@@ -281,3 +320,62 @@ def _check_scalar_valued(function, theta0: np.ndarray, i: int, n_workers: int):
...
+    gi = np.asarray(gi, dtype=float).reshape(-1)  # ensure (m,)
+    return gi
+
+
+def _grad_for_param(
+    function: Callable[[ArrayLike], ArrayLike | float],
+    theta0: ArrayLike,
+    j: int,
+) -> NDArray[np.floating]:
+    """Derivative of a vector-valued function wrt a single parameter theta[j].
```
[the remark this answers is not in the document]
nikosarcevic: yeah exactly. i have removed the ggrad_wrt_param already  fd6442716de1a0ec0d653b52b7e248d3483b69a1

src/derivkit/forecasting/calculus.py
```diff
@@ -281,3 +320,62 @@ def _check_scalar_valued(function, theta0: np.ndarray, i: int, n_workers: int):
...
+
+    Helper used by ``jacobian``. Wraps ``function`` into a single-variable
+    callable via ``derivkit.utils.get_partial_function`` and differentiates it
+    with ``DerivativeKit.adaptive.differentiate``.
```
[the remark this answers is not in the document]
nikosarcevic: also fxied this one c73b015d1e919d992a3915e6742baaa636f1d4d4

src/derivkit/forecasting/calculus.py
```diff
@@ -46,92 +58,85 @@ def gradient(function, theta0, n_workers=1):
...
-            A 1D array or list of parameter values matching the expected
-            input of the function.
-        i: Zero-based index of the parameter with respect to which to differentiate.
-        n_workers: Number of workers used inside
-            ``DerivativeKit.adaptive.differentiate``. This does not parallelize
-            across parameters.
+        function: A function that returns a single value.
+        theta0: The parameter values where the derivative is evaluated.
+        i: The index of the parameter being varied.
+        n_workers: Number of workers used for the internal derivative step.
```
lonbar: The types were specified here because they weren't specified in the function signature. We can do either but we must do one.

tests/test_calculus_jacobian.py
nikosarcevic: good point actually -- updated the test to expect only fp error for non-finite outputs and added a separate test ensuring scalar outputs raise  typerror 

also slightly modified the build jacobian to enure we are kosher

all done in c1d1709e59eb9abb96a6a8ecbba0234c4b1c147b

src/derivkit/forecasting/calculus.py
```diff
@@ -73,33 +85,60 @@ def _grad_component(
...
 
     Returns:
-        (``np.array``): 2D array representing the jacobian.
+        A 2D array representing the jacobian. Each column corresponds to
+            the derivative with respect to one parameter.
+
+    Raises:
+        FloatingPointError: If non-finite values are encountered.
+        ValueError: If ``theta0`` is an empty array.
+        TypeError: If ``function`` does not return a vector value.
```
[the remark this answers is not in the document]
nikosarcevic: fixed c1d1709e59eb9abb96a6a8ecbba0234c4b1c147b

tests/test_calculus_jacobian.py
nikosarcevic: ah good one ------ removed the fixed worker assumptio nso he test now uses 2 workers and skips if the environment can’t spawn extra threads (i wanna actually make this into an issue and perhaps we cna have it package wide. pinning it to my notes)

4163f5eb17e07aa5bbad4d3a44cc41d62aae44e5

src/derivkit/forecasting/calculus.py
nikosarcevic: i tihn kthis is now outdated, can you confirm=?

lonbar: This seems fine now.

PR #152, opened 2025-10-15
Parallelize _get_derivatives; normalize n_workers; avoid nested pools

### LINKED ISSUE #150: Add real parallelization in `_get_derivatives` loops
Right now, ´_get_derivatives()´ runs the loops over parameters one by one, so n_workers doesn’t actually speed things up.

We should add real parallelization so each parameter (for first order) or each parameter pair (for second order) can be processed at the same time using something like ´concurrent.futures´. The goal is to keep results identical to the current output but make it faster when multiple cores are available. It should still default to single-core mode and maintain the same shapes and ordering.

nikosarcevic:
This PR adds real per-parameter/threaded parallelism to `LikelihoodExpansion._get_derivatives` and hardens worker handling.

**Highlights**
- Threaded map across parameters (order=1) and rows m1 (order=2); preserves ordering.
- Normalize `n_workers` (handles None/negatives; caps to cores/params).
- Prevent nested parallelism (inner adaptive `n_workers=1`) → no fork() warnings.
- Refactor to `utils.invert_covariance` with clear diagnostics.
- Tests: equivalence across worker counts, ordering preserved, edge cases.

**Notes**
- No public API changes.
- Numerical results unchanged vs serial.
- Docs/ruff/pytest all green locally.

Closes #150.


lonbar (approved):
LGTM :tada:

PR #157, opened 2025-10-16
Tests/gradient

### LINKED ISSUE #134: unit tests for gradient function in calculus.py
we also need to add unit tests for gradient method in forecasting/calculus.py

this will build on top of #133 (
so i am building on top of:
- we have hessian and jacobian
- both jacobian and hessian unit tests
- bug fix of jacobian which was done in #129 
- and finally pep8 cleanup i did in #133

nikosarcevic:
this pr adds unit tests for gradient method from calculus.py

this builds upon #156 and closes #134  and #160


nikosarcevic:
> does this branch need to be rebase to the main branch?

no i have built on top of tests/hessian

PR #164, opened 2025-10-17
Add poissonian likelihoods

### LINKED ISSUE #74: Implement Poissonian likelihoods

lonbar:
This adds Poissonian likelihoods to forecasts.likelihoods.py. Resolves #74.

My assumption is that the user is interested in calculating probabilities in a way comparable to (7) of <https://arxiv.org/abs/2210.11093>. That is, the user provides the input on a grid of bins (e.g. magnitudes, redshift). I have assumed that this grid is 2D (as in the reference); in the future we could perhaps support more dimensions if there is a need for it.

The difference between the Gaussian and Poissonian cases is that the former has two model inputs (`model_parameters` and `cov`), while the latter has only one (`model_parameters`). I feel that technically `cov` should be part of `model_parameters` in some way (I already brought this up during [discussion of the Gaussian likelihoods](https://github.com/derivkit/derivkit/pull/140#issuecomment-3369101593)). However, I am not entirely sure what form this should take. We could define a data class `ModelParameters`, where these are attributes, or something similar.

Since I have yet to implement an interface for the various likelihoods, perhaps that gives us a clue how to go about this. The data class could house the model parameters and the name of the analysis. This could then be passed to the likelihood interface, which will then return the appropriate data products to the user. Happy to discuss this here or elsewhere.

### DIFF
--- a/tests/test_likelihoods.py
+++ b/tests/test_likelihoods.py
@@ -172,7 +172,7 @@ def test_gaussian_likelihood():
   ]
 )
 
-def test_poissonian_likelihood_negative_parameters(
+def test_poissonian_likelihood_output(
     input_data,
     input_parameters,
     expected_output_data,
@@ -202,7 +202,7 @@ def test_poissonian_likelihood_negative_parameters(
     ([[1, 2, 3], [4, 5, 6]], [[0.1, 0.2, 0.3], [0.4, np.nan, 0.6]]),
   ]
 )
-def test_poissonian_likelihood_incompatible_shapes(test_data, test_parameters):
+def test_poissonian_likelihood_negative_parameters(test_data, test_parameters):
     """Tests negative parameter value exception triggers."""
     with pytest.raises(ValueError):
         dkl.build_poissonian_likelihood(test_data, test_parameters)
@@ -215,7 +215,7 @@ def test_poissonian_likelihood_incompatible_shapes(test_data, test_parameters):
     ([1, 2, 3, 4, 5], [1, 2]),
   ]
 )
-def test_poissonian_likelihood_raises(test_data, test_parameters):
+def test_poissonian_likelihood_incompatible_shapes(test_data, test_parameters):
     """Tests build_poissonian_likelihood shape mismatch exception triggers."""
     with pytest.raises(ValueError):
         dkl.build_poissonian_likelihood(test_data, test_parameters)


lonbar:
@nikosarcevic I have added you as a reviewer. Perhaps @ctrendafilova wants to weight in too (or perhaps do the reviewing)?

src/derivkit/forecasting/likelihoods.py
```diff
@@ -98,9 +98,155 @@ def build_gaussian_likelihood(
...
+    2D row-major array. Similarly, it is possible to supply a single value to
+    ``model_parameters``. The output will still be a 2D array. See Examples for
+    more details.
+
+    Args:
+        data: an array representing the given data values.
+        model_parameters: an array representing the means of the data samples.
+
+    Returns:
+        ``np.ndarray``: an array containing the probability intervals
```
nikosarcevic: should this type hint here? i tihn kwe usually do tuple and say tuple of what

src/derivkit/forecasting/likelihoods.py
```diff
@@ -98,9 +98,155 @@ def build_gaussian_likelihood(
...
+    try:
+        number_events = values_to_reshape.reshape(-1, *parameters.shape[-2:])
+    except ValueError:
+        raise ValueError(
+            "data cannot be reshaped to align with model_parameters: "
+            f"data.shape={values_to_reshape.shape} is incompatible with "
+            f"model_parametres.shape={parameters.shape}."
+        )
+
+    return number_events, poisson.pmf(number_events, parameters)
```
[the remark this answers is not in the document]
nikosarcevic: what i mean is this from spicy

poisson.pmf(k, mu, loc) there are 3 args

src/derivkit/forecasting/likelihoods.py
```diff
@@ -98,9 +98,155 @@ def build_gaussian_likelihood(
...
+
+    Args:
+        data: an array representing the given data values.
+        model_parameters: an array representing the means of the data samples.
+
+    Returns:
+        ``np.ndarray``: an array containing the probability intervals
+            constructed from the data.
+        ``np.ndarray``: an array containing the values of the Poissonian
+            probibility mass function on the probability intervals.
```
lonbar:
```suggestion
        A tuple of arrays containing (in order):
            
            - the data, reshaped to align with the model parameters.
            - the values of the Poissonian probability mass function computed
              from the data and model parameters.
```

src/derivkit/forecasting/likelihoods.py
lonbar: How about the wording in 9ea2d7fcb31f9394eeb53e27a0fcca9e85325c12? I noticed a few typos and inaccuracies while rewriting this, so just rewrote the whole thing.

nikosarcevic: LGTM

src/derivkit/forecasting/likelihoods.py
```diff
@@ -98,9 +98,156 @@ def build_gaussian_likelihood(
...
+    The assumption is that ``model_parameters`` contains the expectation value
+    of some quantity distributed across a grid of bins. The function will try
+    to reshape ``data`` to align with this grid. In principle the grid can
+    contain any number of axes, but currently the number of axes is hardcoded
+    to 2. Supplying a higher-dimensional array to ``model_parameters`` may
+    produce unexpected results.
+
+    This hardcoded limit means that, while it is possible to supply
+    ``model_parameters`` along a 1D grid, the output shape will always be a
+    2D row-major array. Similarly, it is possible to supply a single value to
```
nikosarcevic: is this true?

usay the output is “always 2D” but when model_parameters is scalar your reshape(-1, *parameters.shape[-2:]) makes it 1D. ethier a) don’t promise 2d or b) np.atleast_2d the result consistently

src/derivkit/forecasting/likelihoods.py
```diff
@@ -98,9 +98,156 @@ def build_gaussian_likelihood(
...
+    if np.any(parameters < 0):
+        raise ValueError("values of model_parameters must be non-negative.")
+
+    try:
+        number_events = values_to_reshape.reshape(-1, *parameters.shape[-2:])
+    except ValueError:
+        raise ValueError(
+            "data cannot be reshaped to align with model_parameters: "
+            f"data.shape={values_to_reshape.shape} is incompatible with "
+            f"model_parametres.shape={parameters.shape}."
```
nikosarcevic: typo in model_parameters

src/derivkit/forecasting/likelihoods.py
nikosarcevic: typo

src/derivkit/forecasting/likelihoods.py
```diff
@@ -98,9 +98,156 @@ def build_gaussian_likelihood(
...
+        )
+
```
nikosarcevic: i tihnk we cna also consifrr return_log=True to avoid underflow common with large k

src/derivkit/forecasting/likelihoods.py
```diff
@@ -98,9 +98,155 @@ def build_gaussian_likelihood(
...
+    more details.
+
+    Args:
+        data: an array representing the given data values.
+        model_parameters: an array representing the means of the data samples.
+
+    Returns:
+        ``np.ndarray``: an array containing the probability intervals
+            constructed from the data.
+        ``np.ndarray``: an array containing the values of the Poissonian
```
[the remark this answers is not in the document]
lonbar: I suppose we don't. I rephrased it.

src/derivkit/forecasting/likelihoods.py
lonbar: The precision of `numpy.float64` is about `1e-16`, if I'm not mistaken. I think we can assume that users will be interested in using `numpy.float64`. Are probabilities smaller than this still significant?

nikosarcevic (approved):
I will raise two issues that need to be addressed after this merge

Issue #169, opened 2025-10-19
Gaussian: flexible cov input


nikosarcevic:
Scope: accept cov as scalar, 1D diag vector, or 2D full; validate, symmetrize tiny asymmetry.

Acceptance:

- cov=2.0 → 2.0 * I_k
- cov=np.array([v1,...,vk]) → diag
- cov.shape==(k,k) accepted; non-finite rejected; asymmetric auto-symmetrized (≤1e-12).

Tests: scalar, diag, full, asymmetry small vs large (large should raise).

Branch: feature/gauss-cov-flex

Commit msg template: feat(gauss): support scalar/diag/full covariance with validation

nikosarcevic:
ok gotcha

PR #173, opened 2025-10-19
Add flexible covariance handling in Gaussian likelihood

nikosarcevic:
This PR extends build_gaussian_likelihood to accept scalar, 1D diagonal, or full 2D covariance matrices.

Includes validation, automatic symmetrization of tiny asymmetries, and rejects non-finite values.

Added corresponding unit tests and cleaned up linter/type-hint issues.

This solves #169

### DIFF
--- a/tests/test_likelihoods.py
+++ b/tests/test_likelihoods.py
@@ -36,7 +36,7 @@ def test_gaussian_likelihood_accepts_scalar_diag_full_cov():
     data = np.array([[1.0, 2.0, 3.0], [4.0, 3.0, 1.0]])
     mu = np.array([1.0, 2.0])
 
-    # A scalar covariance should be interpreted as σ² times the identity matrix
+    # A scalar covariance should be interpreted as a multiple of the identity matrix
     grids, pdf = dkl.build_gaussian_likelihood(data, mu, 2.0)
     assert isinstance(grids, tuple) and len(grids) == 2
     assert pdf.shape == (data.shape[1], data.shape[1])


src/derivkit/forecasting/likelihoods.py
```diff
@@ -19,54 +22,56 @@ def build_gaussian_likelihood(
...
-            >>> model_parameters = np.array([0, 4])
-            >>> cov = np.array([[1, 0.2], [0.2, 0.3]])
+            >>> import numpy as np
+            >>> import matplotlib.pyplot as plt
+            >>> data = np.asarray((np.linspace(-10, 10, 30), np.linspace(3, 6, 30)))
+            >>> model_parameters = np.array([0.0, 4.0])
+            >>> cov = np.array([[1.0, 0.2], [0.2, 0.3]])
             >>> grid, pdf = build_gaussian_likelihood(data, model_parameters, cov)
-            >>> plt.contour(*grid, pdf)
+            >>> plt.contour(*grid, pdf)  # doctest: +SKIP
```
lonbar: Don't we want to include this?

src/derivkit/forecasting/likelihoods.py
```diff
@@ -19,54 +22,56 @@ def build_gaussian_likelihood(
...
+            to ensure compatibility with the data and model_parameters.
 
     Returns:
-        ``np.ndarray``: an array containing the coordinate grids constructed
-            from the data.
-        ``np.ndarray``: an array containing the values of the Gaussian
-            likelihood function on the coordinate grid.
+        A tuple:
+          - coordinate_grids: tuple of ndarrays from meshgrid (one per axis).
+          - pdf: ndarray of the Gaussian PDF evaluated on the grid.
```
lonbar: There are some implementation details here. I wouldn't refer to meshgrid; the user needs to know what `coordinate_grids` is, not what was used to construct it. Personally I think the phrasing in the original was fine.

Instead of "pdf" I would say something like probability density function, probabilities, or something similar. "PDF" by itself is too dependent on context.

src/derivkit/forecasting/likelihoods.py
```diff
@@ -78,14 +83,11 @@ def build_gaussian_likelihood(
...
-    elif cov.shape != square_shape:
-        raise ValueError(
-            "cov must be a square 2D array of shape "
-            f"{square_shape}. Actual shape {cov.shape}."
-        )
+    if not np.isfinite(model_parameters).all():
+        raise ValueError("model_parameters contain non-finite values.")
+
+    # Here we validate and normalize the covariance matrix
+    sigma = normalize_covariance(cov, n_parameters=number_model_parameters)
```
lonbar: I don't think an additional function is needed here.

- A scalar is automatically converted to  a multiple of the identity, per the [scipy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.multivariate_normal.html#scipy.stats.multivariate_normal).
-  If `cov` is !D this function is basically just a call to `np.diag`.
- If it is 2D the only thing that really needs to be checked is that the dimensions match with `data`. Scipy will handle any errors.

Adding an additional function may be unnecessary, because we don't know yet if we will need the same logic elsewehere.

tests/test_likelihoods.py
```diff
@@ -7,50 +7,89 @@
 import derivkit.forecasting.likelihoods as dkl
 
 
-def test_gaussian_likelihood():
-    """Test that build_gaussian_likelihood handles input and output correctly."""
-    # Check that 3D data raises ValueError
+def test_gaussian_likelihood_basic_shape_checks():
+    """Check that build_gaussian_likelihood raises exceptions for bad input shapes."""
+    # 3D data -> ValueError
```
lonbar: Please write comments using correct English and full sentences.

tests/test_likelihoods.py
```diff
@@ -7,50 +7,89 @@
...
     with pytest.raises(ValueError):
         data = np.ones((1, 1, 1))
         model_parameter = np.ones(1)
         cov = np.ones((1, 1))
         dkl.build_gaussian_likelihood(data, model_parameter, cov)
 
-    # Check that 2D model_parameter raises ValueError
+    # 2D model_parameter -> ValueError
```
lonbar: Please write in correct English using full sentences

tests/test_likelihoods.py
```diff
@@ -7,50 +7,89 @@
...
-        model_parameter = np.ones((1))
+        model_parameter = np.ones(1)
         cov = np.ones((1, 1))
         dkl.build_gaussian_likelihood(data, model_parameter, cov)
 
-    # Check that cov must be square with axes length equal to model_parameters
+
+def test_gaussian_likelihood_accepts_scalar_diag_full_cov():
+    """Test that build_gaussian_likelihood works for scalar, diag, and full covariances."""
+    # simple 2D grid
```
lonbar:
```suggestion
```
This comment adds no extra information.

tests/test_likelihoods.py
```diff
@@ -7,50 +7,89 @@
...
     with pytest.raises(ValueError):
-        data = np.ones((3, 10))
-        model_parameter = np.ones((3))
-        cov = np.ones((2, 2))
-        dkl.build_gaussian_likelihood(data, model_parameter, cov)
+        dkl.build_gaussian_likelihood(data, mu, np.array([[np.inf]]))
+
+
+def test_gaussian_likelihood_asymmetry_handling():
+    """Test that build_gaussian_likelihood handles asymmetric covariances correctly."""
```
lonbar:
```suggestion
    """Tests that build_gaussian_likelihood handles asymmetric covariances correctly."""
```

tests/test_likelihoods.py
```diff
@@ -7,50 +7,89 @@
...
+    data = np.array([[-1.0, 0.0, 1.0], [3.5, 4.0, 4.5]])
+    mu = np.array([0.0, 4.0])
 
-    # Check that function outputs have the correct types
+    # tiny asymmetry (below 1e-12 relative) -> OK
```
lonbar: Why would small asymmetries be allowed?

src/derivkit/forecasting/likelihoods.py
nikosarcevic: yes this is for pytest to skip the plots part (at least it is what ive seen https://stackoverflow.com/questions/1809037/python-doctest-skip-entire-block)

src/derivkit/forecasting/likelihoods.py
```diff
@@ -19,54 +22,56 @@ def build_gaussian_likelihood(
...
-            f"data must be a 2D array, but is a {data.ndim}D array."
-        )
+        raise ValueError(f"data must be a 2D array, but is a {_data.ndim}D array.")
+
+    # Validate for correct shapes
+    number_samples = _data.shape[0]
+
+    # Validate model parameters
+    model_parameters = np.asarray(model_parameters, dtype=float).ravel()
+    number_model_parameters = model_parameters.shape[0]
```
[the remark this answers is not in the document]
nikosarcevic: 046aba9248db5c7382e689b669474391680407fd

ive cleaned up the shape handilng and removed the unreachable code. the number_samples / model_parameters validation is now grouped together and clearer

ive  initially accepted your suggestion (your scalar change), but I’ll revert it since I realized scalars make no sense here — they have no sample or coordinate axis and break the axis-0 = sample, axis-1 = coordinate convention.

src/derivkit/forecasting/likelihoods.py
nikosarcevic: 85440ce304b25a6b0d9d072812865e733a8828c6

 updated the docstring to describe the meaning of the returned values instead of mentioning meshgrid, and replaced “PDF” with probability density

src/derivkit/forecasting/likelihoods.py
```diff
@@ -78,14 +83,11 @@ def build_gaussian_likelihood(
...
-    square_shape = (number_samples, number_model_parameters)
-    if cov.ndim == 1:
-        cov = cov.reshape(square_shape)
-    elif cov.shape != square_shape:
-        raise ValueError(
-            "cov must be a square 2D array of shape "
-            f"{square_shape}. Actual shape {cov.shape}."
-        )
+    if not np.isfinite(model_parameters).all():
+        raise ValueError("model_parameters contain non-finite values.")
```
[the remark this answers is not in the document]
nikosarcevic: good piint; added here a24aae2e32606ac8f2ca61b7382233001ffca6ff

tests/test_likelihoods.py
```diff
@@ -7,50 +7,89 @@
 import derivkit.forecasting.likelihoods as dkl
 
 
-def test_gaussian_likelihood():
-    """Test that build_gaussian_likelihood handles input and output correctly."""
-    # Check that 3D data raises ValueError
+def test_gaussian_likelihood_basic_shape_checks():
+    """Check that build_gaussian_likelihood raises exceptions for bad input shapes."""
```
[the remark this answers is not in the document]
nikosarcevic: fixed ffa0b05cf2b521c14b5f5955d4a00981330d928c

tests/test_likelihoods.py
nikosarcevic: fixed in 6a8e4372e8065f66ae41189c3358ae3e8eee80e0

tests/test_likelihoods.py
nikosarcevic: diff changed; fixed in 8014076c380d4dba5c4893bb0cb80b2f3a2ede4e

tests/test_likelihoods.py
```diff
@@ -7,50 +7,89 @@
...
+    data = np.array([[-1.0, 0.0, 1.0], [3.5, 4.0, 4.5]])
+    mu = np.array([0.0, 4.0])
 
-    # Check that function outputs have the correct types
+    # tiny asymmetry (below 1e-12 relative) -> OK
+    a_tiny = np.array([[1.0, 1e-14], [0.0, 2.0]])
+    grids, pdf = dkl.build_gaussian_likelihood(data, mu, a_tiny)
+    assert np.isfinite(pdf).all()
+
+    # larger asymmetry -> ValueError
```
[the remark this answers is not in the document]
nikosarcevic: removed in 68173c1a18576d81436d1f117cbc920e2b4ccd1b

src/derivkit/forecasting/likelihoods.py
lonbar: But these are directives for doctest, not pytest. We could use [doctest](https://docs.python.org/3/library/doctest.html), but I think we should add this in a separate PR.

nikosarcevic: you can leave it here and later we can add doctest for sure in a separate pr. jsut leave it in. my pycharm was wraning me about it so i added it

nikosarcevic: the features of these coding suites are fantastic and it helps me a lot make sure everything is kosher. your comment comes off harsh and dismissive.

src/derivkit/forecasting/likelihoods.py
nikosarcevic: I feel strngly to leave it in utils. all cov manipulation should happen there.

PR #188, opened 2025-10-28
refactor(calculus): replace @dataclass with explicit __init__ in CalculusKit; clean imports & docs

### LINKED ISSUE #187: Refactor CalculusKit: replace @dataclass with explicit __init__ constructor
CalculusKit currently uses @dataclass (from the last PR #178 ), which is better suited for simple data containers. Since this class wraps derivative computations (gradient, jacobian, hessian), it should use an explicit __init__ for clarity and flexibility.

Using a dataclass here causes Sphinx autodoc issues and hides initialization logic, offering no real benefit for a computational API.

Proposed fix: replace the dataclass with a standard constructor:
```
class CalculusKit:
    def __init__(self, function, x0):
        self.function = function
        self.x0 = x0
```

nikosarcevic:
refactors CalculusKit to use an explicit __init__ instead of @dataclass.
Makes it the main user-facing api for gradient, Jacobian, and Hessian computations, while keeping derivkit.calculus as a toolbox of pure functions.

passes linting and pytest (except the usual ruff screaming about match and yield.

@lonbar this needs to be done BEFORE the other active PR #185 

this closes #187

### DIFF
--- a/src/derivkit/calculus_kit.py
+++ b/src/derivkit/calculus_kit.py
@@ -49,14 +49,14 @@ class CalculusKit:
         self.function = function
         self.x0 = np.asarray(x0, dtype=float)
 
-    def gradient(self, *, n_workers: int = 1) -> NDArray[np.floating]:
+    def gradient(self, *args, **kwargs) -> NDArray[np.floating]:
         """Returns the gradient of a scalar-valued function."""
-        return build_gradient(self.function, self.x0, n_workers=n_workers)
+        return build_gradient(self.function, self.x0, *args, **kwargs)
 
-    def jacobian(self, *, n_workers: int = 1) -> NDArray[np.floating]:
+    def jacobian(self, *args, **kwargs) -> NDArray[np.floating]:
         """Returns the Jacobian of a vector-valued function."""
-        return build_jacobian(self.function, self.x0, n_workers=n_workers)
+        return build_jacobian(self.function, self.x0, *args, **kwargs)
 
-    def hessian(self, *, n_workers: int = 1) -> NDArray[np.floating]:
+    def hessian(self, *args, **kwargs) -> NDArray[np.floating]:
         """Returns the Hessian of a scalar-valued function."""
-        return build_hessian(self.function, self.x0, n_workers=n_workers)
+        return build_hessian(self.function, self.x0, *args, **kwargs)


src/derivkit/__init__.py
```diff
@@ -1,31 +1,17 @@
 """Provides all derivkit methods."""
 
 from derivkit.adaptive.adaptive_fit import AdaptiveFitDerivative
+from derivkit.calculus_kit import CalculusKit  # <-- the KIT is public here
```
lonbar:
```suggestion
from derivkit.calculus_kit import CalculusKit
```

src/derivkit/calculus_kit.py
```diff
@@ -1,64 +1,64 @@
...
->>>  calc  = CalculusKit(lambda x: x, np.array([1, 2]))
->>>  jacobian = calc.jacobian()
-
-Calculating the Hessian of a scalar-valued function:
->>>  import numpy as np
->>>  from derivkit.calculus_kit import CalculusKit
->>>  calc  = CalculusKit(lambda x: x[0] * x[1], np.array([1, 2]))
->>>  hessian = calc.hessian()
+>>> import numpy as np
+>>> from derivkit.calculus_kit import CalculusKit  # noqa: F401
```
lonbar:
```suggestion
>>> from derivkit.calculus_kit import CalculusKit
```

src/derivkit/calculus_kit.py
```diff
@@ -1,64 +1,64 @@
...
-        """Wrapper function for build_hessian."""
-        return build_hessian(self.function, self.x0, *args, **kwargs)
+    """Provides access to gradient, Jacobian, and Hessian tensors."""
+
+    def __init__(
+        self,
+        function: Callable[[Sequence[float] | np.ndarray], float | NDArray[np.floating]],
+        x0: Sequence[float] | np.ndarray,
+    ):
+        """Initialise with function and expansion point.
```
lonbar:
```suggestion
        """Initialises class with function and expansion point.
```

src/derivkit/calculus_kit.py
```diff
@@ -1,64 +1,64 @@
...
+        """Returns the gradient of a scalar-valued function."""
+        return build_gradient(self.function, self.x0, n_workers=n_workers)
+
+    def jacobian(self, *, n_workers: int = 1) -> NDArray[np.floating]:
+        """Returns the Jacobian of a vector-valued function."""
+        return build_jacobian(self.function, self.x0, n_workers=n_workers)
+
+    def hessian(self, *, n_workers: int = 1) -> NDArray[np.floating]:
+        """Returns the Hessian of a scalar-valued function."""
+        return build_hessian(self.function, self.x0, n_workers=n_workers)
```
lonbar: Please revert these functions to what they were. This wrapper is supposed to be minimal, so it can just pass all arguments it's given to the wrapped functions (`build_hessian` etc). These wrapped functions can then do the argument handling.

nikosarcevic: (I have commited all your previous suggestions)

PR #189, opened 2025-10-28
Refactor/utils package

### LINKED ISSUE #179: Rethink how to handle utils
currently, utils is just one catch-all file with all sorts of things in it. Majority is cov inversion related but there are some random stuff in it. I would suggest to create a directory utils and then have separate scripts for cov inversion/handling and other files we might need. This sohuld not be API visible (meaning that the user should not be albe to see these functions and use/call them).

Thoughts?

nikosarcevic:
What was done in this PR:

- Refactor utils.py into a proper package with clear submodules:
- linalg – public (invert_covariance, normalize_covariance, solve_or_pinv)
- numerics, sandbox, validate – internal helpers
- _concurrency – placeholder for future worker-pool policy
- Remove unused helpers (log_debug_message, is_symmetric_grid, central-difference heuristic).
- Update imports across calculus; no behavior changes.

Why

- Clarifies public vs. internal utilities.
- Prepares for upcoming multi-method differentiate(...) and worker management.

**PLS NOTE:**
_concurrency.py is kept for future use (outer/inner worker coordination).
All tests pass after refactor.

**THIS MUST BE DONE** and merged before #185 !!!! and after #188 as this builds on top of #188 @lonbar !!

this closes #179

### DIFF
--- a/src/derivkit/utils/linalg.py
+++ b/src/derivkit/utils/linalg.py
@@ -1,4 +1,13 @@
-"""Linear algebra helper functions with diagnostics and canonicalization."""
+"""This module provides small linear-algebra helpers with two goals:
+
+1) Diagnostics: warn about non-symmetric inputs, ill-conditioning, and rank issues,
+   and choose a safe fallback when a fast path fails.
+
+2) Canonicalization: accept covariance inputs in multiple forms (scalar, 1D diagonal
+   vector, or 2D matrix) and convert them into a consistent 2D array with validated
+   shape and finite values. In other words, we normalize the input representation
+   so downstream code always receives a well-formed (k x k) array.
+"""
 
 from __future__ import annotations
 


src/derivkit/utils/linalg.py
```diff
@@ -1,280 +1,18 @@
-"""Lightweight utility functions used across DerivKit.
-
-These helpers have no heavy dependencies or side effects and are safe to import
-from anywhere (library code, tests, notebooks). They cover small conveniences
-for logging, quick sanity checks, simple finite-difference heuristics, grid
-symmetry checks, and example/test function generators.
-"""
+"""Linear algebra helper functions with diagnostics and canonicalization."""
```
lonbar: Could you clarify what you mean with "canonicalization"?

src/derivkit/utils/linalg.py
```diff
@@ -426,29 +164,74 @@ def normalize_covariance(
     raise ValueError("cov must be scalar, 1D diag vector, or 2D (k,k) matrix.")
 
 
-def _check_scalar_valued(function, theta0: np.ndarray, i: int, n_workers: int):
-    """Helper used by ``build_gradient`` and ``build_hessian``.
+def solve_or_pinv(matrix: np.ndarray, vector: np.ndarray, *, rcond: float = 1e-12,
```
lonbar: This function is copied from line 204 in the old version of the file, right?

src/derivkit/utils/linalg.py
nikosarcevic: I meant converting different input forms (scalar, vector, or matrix) into a consistent 2D array with proper shape and checks.
Do you want me to write it down explicitly like this?

nikosarcevic:
> > Looks good. Some questions for clarification. I'd like to rebase this after #188 has been merged so we have a better look at the changes in this PR.
> 
> I have now rebased this branch off of the latest commits from the refactor/calculus-init

but of course we should still do it again once the other one is merged

tests/test_utils.py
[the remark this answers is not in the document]
lonbar: We can leave it if you prefer.

src/derivkit/utils/linalg.py
nikosarcevic: smth like this d8739a59be2f6f78a44e6939cd04922c2ae5665b ?

lonbar (approved):

Issue #119, opened 2025-09-29
Update expansions to use calculus functions once calculus is complete


ctrendafilova:
Once the calculus functions are completed, the methods in expansions can be modified to use the calculus functions directly. This will be updated after calculus.py initial implementation is done.

PR #209, opened 2025-11-02
Fixed kwargs passing in hessian.py; replaced expansions' second deriv with build_hessian.

ctrendafilova:
This addresses #119 .

While working on this implementation, I encountered errors due to how `kwargs` was getting passed around throughout `hessian.py`. I have corrected this, I think (at least, I no longer get errors), but please look with a fresh set of eyes and make sure that I have not caused any problems in doing so.

`_row_worker()` and some other methods in `expansions.py` are now vestigial, but I have left the code in place in case it is useful for optimizing `hessian.py`.

### DIFF
--- a/src/derivkit/forecasting/expansions.py
+++ b/src/derivkit/forecasting/expansions.py
@@ -232,7 +232,7 @@ class LikelihoodExpansion:
         if h_raw.shape == (self.n_parameters, self.n_parameters, self.n_observables):
             return h_raw
         raise ValueError(
-            f"build_hessian_tensor returned unexpected shape {h_raw.shape}; "
+            f"build_hessian returned unexpected shape {h_raw.shape}; "
             f"expected ({self.n_observables},{self.n_parameters},{self.n_parameters}) or "
             f"({self.n_parameters},{self.n_parameters},{self.n_observables})."
         )


lonbar:
But we don't want to use the build functions directly, right? We have the `CalculusKit`, which we can use to compute these functions:
```
>>> from derivkit.calculus_kit import CalculusKit
>>> import numpy as np
>>> a = CalculusKit(lambda x: np.cos(x[0])+x[1]*np.exp(-x[2]), [0.3, 3, 1])
>>> a.hessian(method="localpolynomial", n_workers=3)
array([[-9.55336489e-01, -7.83365965e-15,  1.10545502e-14],
       [-7.83365965e-15,  4.03010958e-14, -3.67879418e-01],
       [ 1.10545502e-14, -3.67879418e-01,  1.10363829e+00]])
>>> a.hessian(method="finite", n_workers=3, levels=2)
array([[-9.55336489e-01,  1.33004718e-12, -3.96838118e-12],
       [ 1.33004718e-12, -2.27373675e-13, -3.67879441e-01],
       [-3.96838118e-12, -3.67879441e-01,  1.10363832e+00]])
```
In the argument of `CalculusKit.hessian` etc we should be able to pass all the args and kwargs required for the method used.

Issue #167, opened 2025-10-19
Poisson likelihood: add return_log=True to avoid underflow (use scipy.stats.poisson.logpmf)


nikosarcevic:
Expose a return_log flag so users can compute log-likelihoods directly with poisson.logpmf, avoiding underflow in float64 for large counts / many bins.

poisson.pmf underflows to 0 for tiny probabilities; combining many bins (products or sums of logs) is common in inference.
Working in log-space (sum(log p_i)) is the standard numerically stable approach

change sohuld be very easy to do something like

```
def build_poissonian_likelihood(
    data, model_parameters, return_log: bool = False
) -> tuple[np.ndarray, np.ndarray]:
    ..........
    vals = poisson.logpmf(number_events, params_b) if return_log else poisson.pmf(number_events, params_b)
    return number_events, vals

```
aceptance criteria:

- When return_log=True, outputs match np.log(pmf) within numerical tolerance and never spuriously hit -inf unless mathematically correct.
- Default behavior unchanged (return_log=False).

possible tests

- moderate counts: logpmf ≈ np.log(pmf) (finite, close).
- some edge cases: zeros, large counts, small/large lambdas
- vectorized shapes

PR #210, opened 2025-11-02
Add log-likelihood output to build_poissonian_likelihood

lonbar:
Solves #167.

### DIFF
--- a/src/derivkit/forecasting/likelihoods.py
+++ b/src/derivkit/forecasting/likelihoods.py
@@ -136,7 +136,8 @@ def build_poissonian_likelihood(
     Args:
         data: an array representing the given data values.
         model_parameters: an array representing the means of the data samples.
-        return_log: when set to ``True``, returns the log-likelihood.
+        return_log: when set to ``True``, returns the log-likelihood. Defaults
+            to ``False``.
 
     Returns:
         A tuple of arrays containing (in order):


src/derivkit/forecasting/likelihoods.py
```diff
@@ -135,6 +136,7 @@ def build_poissonian_likelihood(
     Args:
         data: an array representing the given data values.
         model_parameters: an array representing the means of the data samples.
+        return_log: when set to ``True``, return the log-likelihood.
```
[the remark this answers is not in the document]
lonbar: Yes.

lonbar:
```suggestion
        return_log: when set to ``True``, returns the log-likelihood.
```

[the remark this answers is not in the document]
lonbar: Done in 134c890f1fc67f24c4014a11f96701b5e71f3e0a.

src/derivkit/forecasting/likelihoods.py
```diff
@@ -135,6 +136,7 @@ def build_poissonian_likelihood(
     Args:
         data: an array representing the given data values.
         model_parameters: an array representing the means of the data samples.
+        return_log: when set to ``True``, returns the log-likelihood.
```
nikosarcevic: Add what it defaults to

nikosarcevic: i mean it is kinda indicative of what the default is but still would be good to explicitly say defaults to false.

lonbar: The default is already shown in the function signature, though. Sphinx will display that.

nikosarcevic: hmm ok so do we thne need to delete that in all other places? because we explicitly state what default is all over.

nikosarcevic: Ok. I am going to approve this.
In the meantime you might want to raise an issue if we want to remove this everywhere. Your call for me it is fine either way.

nikosarcevic (approved):

Issue #171, opened 2025-10-19
Gaussian: input validation & errors


nikosarcevic:
Scope: clearer checks/messages for shapes & finiteness; enforce len(model_parameters)==k.

Acceptance:

- 1D data promoted; 2D ok; others error with explicit message.
- model_parameters 1D of length k; mismatch errors.
- Finite checks on cov and model_parameters.

Tests: each invalid path raises with informative text.

Branch: chore/gauss-validate-errors

Commit msg: chore(gauss): improve shape/finiteness validation and error messages

PR #213, opened 2025-11-08
Add non-finiteness tests for input in Gauss likelihood

lonbar:
Improves shape/finiteness validation. Solves #171.


lonbar:
@nikosarcevic I think these tests are the only remaining points from #171 which were not yet addressed.

Issue #220, opened 2025-11-10
add Ridders method for derivs


nikosarcevic:
once richardson extrapolation is in, build on top of it and add ridders derivative method

nikosarcevic:
closed with #226

PR #225, opened 2025-11-16
Feature/ridders derivative

### LINKED ISSUE #219: add Richardson extrapolation
add this as a separate method and wire into the finite diff method

nikosarcevic:
added richardson and ridders extension to finite

added unit tests for it (split tests in finite and also in utils into separate files)

I tihnk I had a minor fix in poly (as it was misbehaving)

it is a massive r but in reality it is not conceptually just has bunch of tests

closes #219  and #220 

NOTE: I didnt fix the docs :/

### DIFF
--- a/docs/derivkit.utils.extrapolation.rst
+++ b/docs/derivkit.utils.extrapolation.rst
@@ -1,0 +1,7 @@
+derivkit.utils.extrapolation module
+===================================
+
+.. automodule:: derivkit.utils.extrapolation
+   :members:
+   :show-inheritance:
+   :undoc-members:


src/derivkit/finite/finite_difference.py
```diff
@@ -64,59 +62,79 @@ class FiniteDifferenceDerivative:
...
+        levels: int | None = None,
+        return_error: bool = False,
+    ) -> np.ndarray | float:
         """Computes the derivative using a central finite difference scheme.
 
         Supports 3-, 5-, 7-, or 9-point central difference stencils for
         derivative orders 1 through 4 (depending on the stencil size).
         Derivatives are computed for scalar or vector-valued functions.
+        Allows for optional extrapolation (Richardson or Ridders) to improve accuracy.
+        It also returns an error estimate if requested.
```
lonbar: I thought these methods are distinct applications of finite differences, and that the current class implements 5-point stencils. Shouldn't the Richardson and Ridders methods be included here?

Issue #233, opened 2025-11-18
Issues with passing dk_kwargs to calculus functions


ctrendafilova:
Here is a more detailed description of the issue under discussion in PR #209 .

The goal is to rewrite `LikelihoodExpansion._get_derivatives` so that it directly uses the `CalculusKit` `jacobian` and `hessian` methods, which will greatly simplify the code.

The derivatives module has been restructured such that a flexible, and optional, dictionary of keyword arguments can be passed by the user, to customize the differentiation computation. This dictionary must be passable through the functions of the `LikelihoodExpansion` class so that the differentiation options can be customized here as well.

Here:
https://github.com/derivkit/derivkit/blob/main/src/derivkit/forecasting/expansions.py#L153
the `_get_derivatives` method is defined with an input argument `dk_kwargs`, which is either a dictionary, or `None` type.
This is being passed from `get_forecast_tensors`, which is what we intend for the user to interact with:
https://github.com/derivkit/derivkit/blob/main/src/derivkit/forecasting/expansions.py#L92
The user will be passing a dictionary, which is carried along, until the code calls a calculus function, such as here:
https://github.com/derivkit/derivkit/blob/main/src/derivkit/forecasting/expansions.py#L201
```
                build_jacobian(
                    ... ,
                    dk_kwargs=dk_kwargs,
                )
```

The issue is that doing the analogous thing for the second-order derivatives by directly calling `CalculusKit.hessian` or `build_hessian` is not successful.

The following code will replicate the issue in a standalone way.

```
def myf(paramList):
    x = paramList[0]
    y = paramList[1]
    obs1 = x + y
    obs2 = x + y
    obsList = np.array([obs1,obs2])
    return obsList
```

`dictionary_of_kwargs = {"levels":2}`

`b = CalculusKit(myf, np.array([1, 2]))`

This works:
`b.jacobian(method="finite")`

This works:
`b.jacobian(method="finite",dk_kwargs = None)`

This works:
`b.jacobian(method="finite",dk_kwargs = dictionary_of_kwargs)`

Passing the arguments directly:
`b.jacobian(method="finite", levels=2)`
does not work:
`TypeError: build_jacobian() got an unexpected keyword argument 'levels'`

This works:
`b.hessian(method="finite")`

This:
`b.hessian(method="finite",dk_kwargs = None)`
gives this error:
`TypeError: FiniteDifferenceDerivative.differentiate() got an unexpected keyword argument 'dk_kwargs'`

And this:
[...]

PR #235, opened 2025-11-19
Bringing dk_kwargs in jacobian in line with hessian

ctrendafilova:
This PR attempts to address #233 .

I have changed the usage of `dk_kwargs` throughout `jacobian.py` to match the treatment in `hessian.py`.

I had to make some modifications in `expansions.py` because it currently calls the jacobian directly.

I have NOT replaced the second derivative computation with the hessian. I will do this in a new PR after this one is merged.


lonbar (commented):
Looks good to me. I made a few minor suggestions. Thanks for doing this.

PR #238, opened 2025-11-21
Multi x0 array support

nikosarcevic:
Adds support for array-valued x0 in DerivativeKit and includes full test coverage for multi-point evaluation.

- Loops over x0 arrays and stacks results
- Preserves input shape
- Adds fake-engine tests + real finite-diff, poly and adaptive

### DIFF
--- a/tests/test_derivative_kit.py
+++ b/tests/test_derivative_kit.py
@@ -26,7 +26,7 @@ def _reset_dispatch_state() -> None:
 
 
 class _FakeAdaptiveEngine:
-    """Fake adaptive engine that records ctor args and differentiate kwargs."""
+    """Fake adaptive engine that records constructor args and differentiate kwargs."""
 
     def __init__(self, function, x0, **kwargs):
         """Records ctor arguments."""
@@ -43,7 +43,7 @@ class _FakeAdaptiveEngine:
 
 
 class _FakeFiniteEngine:
-    """Fake finite-difference engine that records ctor args and differentiate kwargs."""
+    """Fake finite-difference engine that records constructor args and differentiate kwargs."""
 
     def __init__(self, function, x0, **kwargs):
         """Records ctor arguments."""
@@ -128,7 +128,7 @@ def test_adaptive_dispatch(monkeypatch):
 
     assert out == 0.0
 
-    # ctor was called for adaptive, not for finite
+    # constructor was called for adaptive, not for finite
     assert calls["adaptive"]  # non-empty dict
     adaptive_call = calls["adaptive"]
     f_called = adaptive_call["function"]
@@ -162,7 +162,7 @@ def test_finite_dispatch(monkeypatch):
 
     assert out == 0.0
 
-    # ctor was called for finite, not for adaptive
+    # constructor was called for finite, not for adaptive
     assert calls["finite"]
     finite_call = calls["finite"]
     f_called = finite_call["function"]


tests/test_derivative_kit.py
```diff
@@ -112,3 +113,89 @@ def test_default_method_is_adaptive():
...
+    def fake_resolve(name: str):
+        return FakeEngine
+
+    monkeypatch.setattr("derivkit.derivative_kit._resolve", fake_resolve, raising=True)
+
+    dk = DerivativeKit(quad, x0=x0_grid)
+    out = dk.differentiate(order=1)
+
+    # shape: x0.shape + (3,)
+    assert out.shape == (2, 2, 3)
```
lonbar: The docstring of `DerivativeEngine` states that the shape should be `(3,2,2)`.

tests/test_derivative_kit.py
```diff
@@ -112,3 +113,89 @@ def test_default_method_is_adaptive():
...
+            return np.array([self.x0, 2 * self.x0])
+
+    def fake_resolve(name: str):
+        # ignore method name; always use FakeEngine
+        return FakeEngine
+
+    monkeypatch.setattr("derivkit.derivative_kit._resolve", fake_resolve, raising=True)
+
+    dk = DerivativeKit(quad, x0=x0_vals)
+    out = dk.differentiate(order=1)  # method=None → default
```
lonbar:
```suggestion
    out = dk.differentiate(order=1)
```

lonbar: Please write comments using full sentences in grammatical English.

tests/test_derivative_kit.py
```diff
@@ -112,3 +113,89 @@ def test_default_method_is_adaptive():
...
+
+def test_array_x0_loops_and_stacks(monkeypatch):
+    """Tests that array x0 values are handled correctly by looping and stacking results."""
+    x0_vals = np.array([-0.5, 0.0, 0.5])
+    seen_x0 = []
+
+    class FakeEngine:
+        def __init__(self, function, x0, **kwargs):
+            # record scalar x0 passed to each engine instance
+            seen_x0.append(x0)
```
lonbar: I feel that the class shouldn't affect the state outside of its scope.

lonbar: This means that changes become non-local which can make debugging difficult.

nikosarcevic: yeah thanks that’s a good point. ive reworked the tests so the fakes no longer mutate external variables captured from their scope. the dispatch tests now record call information in small module-level tracking dicts, and the array-x0 test uses a class attribute (_FakeArrayX0Engine.instances) that is explicitly reset at the start of the test

 that keeps all state local to the helper fakes and makes the effects easier to reason about and debug
 i hope you are ok with it now f5398326a7ae3f464bd021c0d904ab263d343f63

tests/test_derivative_kit.py
```diff
@@ -112,3 +113,89 @@ def test_default_method_is_adaptive():
...
     else:
         assert y_none == y_adpt
+
+
+def test_array_x0_loops_and_stacks(monkeypatch):
+    """Tests that array x0 values are handled correctly by looping and stacking results."""
+    x0_vals = np.array([-0.5, 0.0, 0.5])
+    seen_x0 = []
+
+    class FakeEngine:
```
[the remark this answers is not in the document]
nikosarcevic: fixed in f5398326a7ae3f464bd021c0d904ab263d343f63

tests/test_derivative_kit.py
```diff
@@ -112,3 +113,89 @@ def test_default_method_is_adaptive():
...
+    x0 = np.array([-0.3, 0.0, 0.7])
+    dk = DerivativeKit(f, x0)
+
+    out = dk.differentiate(order=1, method=method, **extra_kwargs)
+    expected = np.cos(x0)
+
+    np.testing.assert_allclose(out, expected, rtol=1e-5, atol=1e-7)
+
+
+def test_array_x0_preserves_shape_for_2d_grid(monkeypatch):
```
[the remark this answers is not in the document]
lonbar: In other words, if the users wants to differentiate a function from some library they first have to write a wrapper and have derivkit operate on that? We don't want to handle directional derivatives of some sort?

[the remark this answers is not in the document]
lonbar: Let's postpone this discussion for now.

Issue #245, opened 2025-11-21
Rename tests (super small fix)


nikosarcevic:
Currently some files in tests just need an additional name in filename to reflect which kit they belong to.

This should not change the tests or behaviour -- only file naming.

Issue #177, opened 2025-10-22
Small fix: reorganize and establish test naming standards


nikosarcevic:
Currently, we do not have an established naming standard for test files.
Decide on the naming conventions and rename the files accordingly.

nikosarcevic:
closed with #246

PR #246, opened 2025-11-21
renamed unit tests

nikosarcevic:
tin this PR i have only renamed the names of the files in tests. no other changes but this (meaning I have not changed the tests or any behaviour). this is only to keep tests clean and easy to find.

solves #245 and #177


lonbar (approved):
I like it. Thanks for doing this.

PR #252, opened 2025-11-21
added tests for forecast_kit/fisher

### LINKED ISSUE #248: Add dedicated Fisher tests
Create a new test file that directly tests the Fisher computation (shape, type, delegation, and argument forwarding). This should not go through the big combined delegation test.

nikosarcevic:
added tests for forecast_it fisher into a separate file.

this closes #248

### DIFF
--- a/tests/test_forecast_kit_fisher.py
+++ b/tests/test_forecast_kit_fisher.py
@@ -66,7 +66,8 @@ def forecasting_mocks(monkeypatch):
 
 
 def test_get_forecast_tensors_order1_matches_matrix_product(forecasting_mocks):
-    """Tests that the Fisher from order=1 equals D1 @ invcov @ D1.T."""
+    """Tests that the Fisher matrix equals the derivative matrix @ invcov."""
+    # derivative order 1
     d1 = np.array(
         [
             [1.0, 2.0],


tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,314 @@
...
+    INVCOV_CALL_INFO = {
+        "cov": cov_arr,
+        "warn_prefix": warn_prefix,
+    }
+    return INVCOV_GLOBAL
+
+
+def two_obs_model(theta):
+    """Model that always returns a 2-element observable vector."""
+    _ = np.atleast_1d(theta)
```
lonbar:
```suggestion
```

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,314 @@
...
+    global D1_GLOBAL, INVCOV_GLOBAL, DERIV_CALL_INFO, INVCOV_CALL_INFO
+    D1_GLOBAL = np.array(
+        [
+            [1.0, 0.5],
+            [-0.3, 2.0],
+        ]
+    )
+    INVCOV_GLOBAL = np.array([[10.0, 0.0], [0.0, 0.5]])
+    DERIV_CALL_INFO = {}
+    INVCOV_CALL_INFO = {}
```
lonbar: THese were already defined in the larger scope, so I don't think it's necessary to define them again.

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,314 @@
...
+import numpy as np
+import pytest
+
+from derivkit.forecasting.expansions import LikelihoodExpansion
+
+# Globals used by fake helpers
+D1_GLOBAL = None
+DERIV_CALL_INFO: dict = {}
+INVCOV_GLOBAL = None
+INVCOV_CALL_INFO: dict = {}
```
lonbar: You define these attributes in the `fake_` functions, so probably you don't need to define these outside of the functions?

lonbar (commented):
I think these tests are okay, but feel that it would be clearer to define `D1_GLOBAL`, `INVCOV_GLOBAL`, etc be defined in the `fake_` functions and their attributes set accordingly. Repeatedly redefining them leaves a lot of different places which have to be kept in mind if this ever changes.

tests/test_forecast_kit_fisher.py
nikosarcevic: I tried dropping the module-level initialisations, but that triggers “undefined global” warnings

tests/test_forecast_kit_fisher.py
nikosarcevic: i modified so it is now calling this helepr wrapper f686dbb8b66bb6ddd060ae0bece72febe14ecd7d

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+        mocks.fake_get_derivatives,
+        raising=True,
+    )
+
+    return mocks
+
+
+def test_build_fisher_matches_matrix_product():
+    """Tests that _build_fisher computes Fisher matrix as expected."""
+    d1 = np.array(
```
lonbar: In general you may want to rename `d1` to `data_vector` or something similar, because the name `d1` doesn't really clarify its use.

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+        "derivkit.forecasting.expansions.LikelihoodExpansion._get_derivatives",
+        mocks.fake_get_derivatives,
+        raising=True,
+    )
+
+    return mocks
+
+
+def test_build_fisher_matches_matrix_product():
+    """Tests that _build_fisher computes Fisher matrix as expected."""
```
lonbar: This may be a bit controversial, but `_build_fisher` is supposed to be an internal helper function. It should not be user facing: it's an implementation detail. We may want to remove it, at which point we also have to remove the test. Also, I'm not entirely sure that testing it makes sense currently, as it contains a single execution of a numpy function (though tbf that may change in the future).

lonbar: Basically, I feel that in this case we want to test the fisher output of `get_forecast_tensors`.

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def test_get_forecast_tensors_order1_builds_fisher(forecasting_mocks):
+    """Tests that get_forecast_tensors order=1 builds Fisher matrix correctly."""
+    theta0 = np.array([0.1, -0.2])
+    cov = np.array([[1.0, 0.2], [0.2, 2.0]])
+
+    forecasting_mocks.set_state(
+        d1=np.array([[1.0, 0.5], [-0.3, 2.0]]),
+        invcov=np.array([[10.0, 0.0], [0.0, 0.5]]),
```
lonbar: This is not the inverse of `cov`.

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+    kwargs = forecasting_mocks.deriv_call_info["kwargs"]
+    assert kwargs["order"] == 1
+    assert kwargs["method"] == method
+    assert kwargs["n_workers"] == 2
+    assert kwargs["extrapolation"] == extrapolation
+    assert kwargs["stencil"] == stencil
+
+
+def test_get_forecast_tensors_order1_forwards_local_polyfit_kwargs(forecasting_mocks):
+    """Test that local_polyfit method and its kwargs are forwarded to _get_derivatives."""
```
lonbar: `_get_derivatives` is an auxiliary function that is not/should not be visible to the user. Is that something we want to test?

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def three_obs_model(theta):
+    """Model that always returns a 3-element observable vector."""
+    return np.zeros(3, dtype=float)
+
+
+def wrong_length_model(theta):
+    """Model that returns a wrong-length observable vector (for shape check)."""
+    return np.zeros(3, dtype=float)
```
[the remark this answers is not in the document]
nikosarcevic: good catch. removed and used three_obs_model instead 2a800bc64ec288cb9c0699c7055889867254d1de

tests/test_forecast_kit_fisher.py
nikosarcevic: I also removed another call to _build_fisher in 78487f9ac3bc0b6c58023d59a6bdd58489c4c244

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+    np.testing.assert_allclose(fisher, expected)
+
+
+def test_build_fisher_is_symmetric():
+    """Tests that _build_fisher returns a symmetric Fisher matrix."""
+    rng = np.random.default_rng(123)
+    d1 = rng.normal(size=(3, 4))
+    a = rng.normal(size=(4, 4))
+    cov = a @ a.T + np.eye(4)
+    invcov = np.linalg.inv(cov)
```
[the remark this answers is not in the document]
nikosarcevic: yes

tests/test_forecast_kit_fisher.py
nikosarcevic: also fixed elsehwere to make sure invcov is kosher bd0aaaa191e6eae7ae72658b7353a0a42a48b019

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+
+    # wrong_length_model returns 3 observables, so this should fail
+    lx = LikelihoodExpansion(function=wrong_length_model, theta0=theta0, cov=cov)
+
+    with pytest.raises(ValueError):
+        lx.get_forecast_tensors(forecast_order=1)
+
+
+def test_get_forecast_tensors_order1_default_n_workers(forecasting_mocks):
+    """Tests that default n_workers=1 is used in _get_derivatives."""
```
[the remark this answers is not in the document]
nikosarcevic: fixed in f48015ea158a00dfff634b85a4ee95b6d31785e6 and 78487f9ac3bc0b6c58023d59a6bdd58489c4c244

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+@pytest.mark.parametrize("method", ["adaptive", "finite"])
+@pytest.mark.parametrize("extrapolation", ["richardson", "ridders", "gauss_richardson"])
+@pytest.mark.parametrize("stencil", [3, 5, 7, 9])
+def test_get_forecast_tensors_order1_forwards_derivative_kwargs(
+    forecasting_mocks,
+    method,
+    extrapolation,
+    stencil,
+):
+    """Tests that derivative method and its kwargs are forwarded to _get_derivatives."""
```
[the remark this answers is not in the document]
nikosarcevic: I think this was fixed  in  f48015ea158a00dfff634b85a4ee95b6d31785e6 and 78487f9ac3bc0b6c58023d59a6bdd58489c4c244

tests/test_forecast_kit_fisher.py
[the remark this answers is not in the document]
nikosarcevic: thisis a simple property test of the public api-- for a symmetric cov, the fisher must be symmetric.
The matrix-product tests cover the exact contraction -- this one just checks the expected high-level behaviour

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+            self.invcov = None
+            self.deriv_call_info = None
+            self.invcov_call_info = None
+
+        def set_state(self, d1, invcov):
+            """Sets the derivative and inverse covariance matrices for this test."""
+            self.d1 = np.asarray(d1, dtype=float)
+            self.invcov = np.asarray(invcov, dtype=float)
+            self.deriv_call_info = None
+            self.invcov_call_info = None
```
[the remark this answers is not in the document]
nikosarcevic: good catch. removed in 79f0843ed6ce31fe70acaa1a2bda8fbcf12d3127

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,276 @@
...
+        mocks.fake_invert_covariance,
+        raising=True,
+    )
+    monkeypatch.setattr(
+        "derivkit.forecasting.expansions.LikelihoodExpansion._get_derivatives",
+        mocks.fake_get_derivatives,
+        raising=True,
+    )
+
+    return mocks
```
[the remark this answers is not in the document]
nikosarcevic: Thanks — good point. I replaced the custom class with a simple state dict and plain funtcions, so the mocks no longer “connect” classes. this keeps the fixture minimal and avoids coupling
5dce0d6af6c9230553f4786544f9402df5a8cbf1

abd here I fixed the docstrings fe5740cae3ed051364f5569af51dfb05e3d41745

tests/test_forecast_kit_fisher.py
```diff
@@ -0,0 +1,289 @@
...
+@pytest.mark.parametrize("method", ["adaptive", "finite"])
+@pytest.mark.parametrize("extrapolation", ["richardson", "ridders", "gauss_richardson"])
+@pytest.mark.parametrize("stencil", [3, 5, 7, 9])
+def test_get_forecast_tensors_order1_forwards_derivative_kwargs(
+    forecasting_mocks,
+    method,
+    extrapolation,
+    stencil,
+):
+    """Tests that derivative method and its kwargs are forwarded to he internal derivative routine."""
```
lonbar:
```suggestion
    """Tests that derivative method and its kwargs are forwarded to the internal derivative routine."""
```

Issue #251, opened 2025-11-21
Add tests for delta_nu


nikosarcevic:
Add a small test ensuring that ForecastKit.delta_nu delegates correctly to LikelihoodExpansion.build_delta_nu and returns the exact value produced by the underlying object.

Issue #250, opened 2025-11-21
Add tests for fisher_bias


nikosarcevic:
Currently fisher_bias has no tests. Add a test file ensuring it correctly forwards arguments to LikelihoodExpansion.build_fisher_bias, including rcond, n_workers, and any dk_kwargs.

nikosarcevic:
closed with #253

PR #253, opened 2025-11-21
added tests for fisher bias and delta_nu

nikosarcevic:
added unit tests for fisher bias and delta nu.

solves #250 and #251 (no need to have a separate file for delta_nu)

### DIFF
--- a/tests/test_forecast_kit_fisher_bias.py
+++ b/tests/test_forecast_kit_fisher_bias.py
@@ -187,69 +187,38 @@ def test_build_delta_nu_exceptions(data_with, data_without, model, cov, expected
         lx.build_delta_nu(data_with=data_with, data_without=data_without)
 
 
-@pytest.mark.parametrize("method", ["adaptive", "finite", "local_polynomial"])
-def test_build_fisher_bias_supports_all_methods(method):
-    """Tests that all supported differentiation methods work without error."""
-    theta0 = np.array([0.1, -0.2])
-    cov = np.diag([2.0, 1.0, 0.5])
-    fisher = np.array([[2.0, 0.1], [0.1, 1.5]])
-    delta_nu = np.array([1.0, 2.0, 3.0])
-
-    lx = LikelihoodExpansion(function=linear_model, theta0=theta0, cov=cov)
-
-    bias_vec, delta_theta = lx.build_fisher_bias(
-        fisher_matrix=fisher,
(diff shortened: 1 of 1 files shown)


tests/test_forecast_kit_fisher_bias.py
```diff
@@ -0,0 +1,525 @@
...
+
+import numpy as np
+import pytest
+
+from derivkit.forecasting.expansions import LikelihoodExpansion
+
+
+def three_obs_model(theta):
+    """Model that always returns a 3-element observable vector."""
+    _ = np.atleast_1d(theta)
```
lonbar: Is this supposed to test something?

tests/test_forecast_kit_fisher_bias.py
```diff
@@ -0,0 +1,327 @@
...
+    return np.zeros(3, dtype=float)
+
+
+def two_obs_model(theta):
+    """Model that always returns a 2-element observable vector."""
+    _ = np.atleast_1d(theta)
+    return np.zeros(2, dtype=float)
```
lonbar: If these are supposed to test something it would be better to to run `assert`. Assigning to `_` doesn't do much in this case.

tests/test_forecast_kit_fisher_bias.py
```diff
@@ -0,0 +1,327 @@
...
+    theta0 = np.array([0.0])
+    cov = np.eye(3)
+
+    lx = LikelihoodExpansion(function=three_obs_model, theta0=theta0, cov=cov)
+
+    a = np.array([0.0, np.nan, 1.0])
+    b = np.zeros(3)
+
+    with pytest.raises(FloatingPointError):
+        lx.build_delta_nu(data_with=a, data_without=b)
```
lonbar: Again duplicates of the same tests, with different input. `a` and `b` are not good variable names.

lonbar (changes requested):
I notice a lot of duplication. It isn't clear why we should distinguish each of these cases if the tests are practically identical, so I would suggest we merge them.

tests/test_forecast_kit_fisher_bias.py
nikosarcevic: reorganized so we dont duplicate checks for diag and non diagnostic in 416db355120a565d6f18bbf286e3c2d43706bd49

lonbar (approved):

PR #259, opened 2025-11-26
Add full Tabulated1DModel support + tests + fixes

### LINKED ISSUE #258: Support tabulated (x, y) data as a DerivativeKit model
Allow DerivativeKit to handle derivatives of tabulated data (x, y) directly, for those instances where a user does not have an actual functional form but just an array.

How we can do this:
Add a small callable wrapper Tabulated1DModel (interpolates y(x)) in a new file tabulated_model.py (or some better name)
Add DerivativeKit.from_tabulated_1d(x, y, x0, ...) which wraps (x, y) into the model internally.
No changes to derivative engines; they work unchanged since they only expect a callable.

Users then can do this basically:
kit = DerivativeKit.from_tabulated_1d(x, y, x0=0.5)
d1 = kit.differentiate(method="finite", order=1)
with full support for all DerivKit derivative methods.

nikosarcevic:
This PR adds complete support for **Tabulated1DModel** across DerivativeKit, CalculusKit, and ForecastKit, including scalar, vector, and tensor outputs.

### **Main changes**

* extended derivkit to have Tabulated1DModel support up to **tensor-valued** observables.
* added  tests:

  * tabulated one_d 
  * gradient, Jacobian, Hessian on tabulated models
  * Fisher + Fisher bias
  * DALI (quadratic + cubic models)
  * C-order flattening behaviour
* Ensured consistent **row-major (C-order)** handling throughout the library 
* Fixed a finite-difference layout issue uncovered during testing (luckily caught it now!!!)
* Removed deprecated numpy scalar conversions (luckily caught it now!!!)

Tabulated models behave like analytic models across all derivative engines and all forecasting functionality, with consistent shapes and ordering.

This closes #258.

Note to Matthijs: @lonbar  -- I did not touch docs. We should also add this to the front page of docs to make sure this is mentioned.

### DIFF
--- a/src/derivkit/calculus/hessian.py
+++ b/src/derivkit/calculus/hessian.py
@@ -397,7 +397,7 @@ def _mixed_partial_value(
         n_workers=n_workers,
         **dk_kwargs,
     )
-    return float(np.asarray(val, dtype=float).item())
+    return float(val)
 
 
 def _build_hessian_internal(


src/derivkit/utils/validate.py
```diff
@@ -62,3 +64,40 @@ def check_scalar_valued(function, theta0: np.ndarray, i: int, n_workers: int):
...
+    x: ArrayLike,
+    y: ArrayLike,
+) -> tuple[NDArray[np.floating], NDArray[np.floating]]:
+    """Validate and convert tabulated (x, y) data into NumPy arrays.
+
+    Requirements:
+      - x is 1D and strictly increasing.
+      - y has at least 1 dimension.
+      - y.shape[0] == x.shape[0], but y may have arbitrary trailing
+        dimensions (scalar, vector, or ND output).
```
lonbar: ### 
```suggestion
      - ``x`` is 1D and strictly increasing.
      - ``y`` has at least 1 dimension.
      - ``y.shape[0] == x.shape[0]``, but ``y`` may have arbitrary trailing
        dimensions (scalar, vector, or ND output).
```

src/derivkit/utils/validate.py
```diff
@@ -62,3 +64,40 @@ def check_scalar_valued(function, theta0: np.ndarray, i: int, n_workers: int):
             "build_gradient() expects a scalar-valued function; "
             f"got shape {probe.shape} from full_function(params)."
         )
+
+
+def validate_tabulated_xy(
+    x: ArrayLike,
+    y: ArrayLike,
+) -> tuple[NDArray[np.floating], NDArray[np.floating]]:
+    """Validate and convert tabulated (x, y) data into NumPy arrays.
```
lonbar: The input isn't a `tuple` as suggested by the docstring

src/derivkit/local_polynomial_derivative/local_polynomial_derivative.py
```diff
@@ -73,6 +73,8 @@ def differentiate(
             diagnostics: If True, returns a diagnostics dictionary along with the derivative estimate.
 
         Returns:
+            If return_error is True and diagnostics is False: A tuple (derivative, error).
+            If return_error and diagnostics are both True: A tuple (derivative, error, diagnostics_dict).
```
lonbar: These are a bit confusing in combination with the two potential cases directly underneath this.

src/derivkit/tabulated_model/one_d.py
```diff
@@ -0,0 +1,236 @@
...
+            fill_value: Value for out-of-bounds evaluation when extrapolation is
+                disabled. If ``None``, a ``ValueError`` is raised instead.
+        """
+        x_arr, y_arr = validate_tabulated_xy(x, y)
+
+        self.x = x_arr
+
+        if y_arr.ndim == 1:
+            self._out_shape: tuple[int, ...] = ()
+            y_flat = y_arr[:, None]
```
lonbar:
```suggestion
            y_flat = y_arr[:, np.newaxis]
```

src/derivkit/tabulated_model/one_d.py
```diff
@@ -0,0 +1,236 @@
...
+        x = arr[:, 0]
+        if n_cols == 2:
+            y = arr[:, 1]
+        else:
+            y = arr[:, 1:]
+        return x, y
+
+    raise ValueError(
+        f"Unexpected table shape {arr.shape}; expected (N, 2), (N, M+1) or (2, N)."
+    )
```
lonbar: I feel it would be clearer if this is a `match`, with a single return at the end.

src/derivkit/tabulated_model/one_d.py
```diff
@@ -0,0 +1,236 @@
...
+        table: 2D array containing x and y columns.
+        extrapolate: Whether to allow evaluation outside the x range.
+        fill_value: Value for out-of-range evaluation when extrapolation is
+            disabled.
+
+    Returns:
+        A :class:`Tabulated1DModel` constructed from the parsed table.
+    """
+    x, y = parse_xy_table(table)
+    return Tabulated1DModel(x, y, extrapolate=extrapolate, fill_value=fill_value)
```
lonbar: This function isn't really necessary; you can have `parse_xy_table` return a `Tabulated1DModel` instead.

tests/test_derivative_kit_output_layout.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def vector_func(x: float) -> np.ndarray:
+    """A 2D vector-valued linear function: f'(x) = [1, 2]."""
+    return np.array([x, 2.0 * x])
+
+
+def tensor_func(x: float) -> np.ndarray:
+    """A tensor-valued linear function (2x2 matrix)."""
+    return np.array([[x, 2.0 * x], [3.0 * x, 4.0 * x]])
```
lonbar: The summary line must start with an action, e.g. "returns a scalar linear function". Please use full sentences and grammatical english.

tests/test_tabulated_one_d_calculus.py
```diff
@@ -0,0 +1,138 @@
...
+    y22 = 4.0 * x_tab
+    y_tab = np.array([[y11, y12], [y21, y22]]).transpose(2, 0, 1)
+    return Tabulated1DModel(x_tab, y_tab)
+
+
+def tensor_observable(theta: np.ndarray) -> np.ndarray:
+    """A tabulated tensor observable: observable = tabulated_tensor(theta[0])."""
+    model = _make_tabulated_linear_tensor()
+    mat = model(theta[0])
+    return np.asarray(mat, dtype=float).ravel(order="C")
```
lonbar: Please write the summary lines in plain English. Why are some private and some public?

tests/test_tabulated_one_d_derivatives.py
```diff
@@ -0,0 +1,111 @@
...
+
+def _make_linear_tensor_model():
+    """Creates a Tabulated1DModel for a 2x2 tensor-valued linear function."""
+    x_tab = np.linspace(-2.0, 2.0, 41)
+    y11 = 2.0 * x_tab + 1.0
+    y12 = -1.0 * x_tab + 0.0
+    y21 = 0.5 * x_tab + 2.0
+    y22 = -3.0 * x_tab + 4.0
+    y_tab = np.array([[y11, y12], [y21, y22]]).transpose(2, 0, 1)
+    return Tabulated1DModel(x_tab, y_tab)
```
lonbar: Why do these have an `_` at the front of their name?

tests/test_tabulated_one_d_forecasting.py
lonbar: It's a bit confusing why some of the auxiliary files have leading underscores and some don't. I think the summary lines that break are a bit messy, and it's better to write more context in an extended docstring. So, instead of e.g.
```
"""Returns a tabulated cubic model: f(x) = x^3."""
```
write something like
```
"""Returns a cubic mononomial of unit weight.

Specifically it returns the value of :math: f(x) = x^3.
"""
or something similar.

File: src/derivkit/utils/validate.py (hunk 2 above)
Comment (nikosarcevic) [reply src/derivkit/utils/validate.py]: yeah good catch. fixed in fee7b703e91d8e66ee0198b67ef8958bcdb88c4a

File: src/derivkit/calculus/hessian.py [hunk 10]
```diff
@@ -291,6 +291,8 @@ def _hessian_component_worker(
         n_workers=inner_workers or 1,
         **dk_kwargs,
     )
+    val_arr = np.asarray(val, dtype=float)
+    return float(val_arr.ravel(order="C")[0])
```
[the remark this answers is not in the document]
nikosarcevic: good catch ---  I originally added the np.asarray(...).ravel(order="C")[0] here when I was making sure all the hass paths used the same C-order flattening, but in this spot _hessian_component is already guaranteed to be scalar, so that’s overkill for real. I’ve simplified it to return float(val) and kept the stricter scalar check in _hessian_component where it belongs

fixed in 5e58a2b71d920b482ba47d5b191789659bf1d737

src/derivkit/finite/core.py
```diff
@@ -60,10 +60,26 @@ def single_finite_step(
         dtype=float,
     )
 
+    # values shape: (n_stencil,) for scalar outputs,
+    # (n_stencil, ...) for vector/tensor outputs.
```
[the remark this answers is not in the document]
nikosarcevic: Yep — the trailing dims are exactly the function’s output shape. I clarified the comment to:
values shape: (n_stencil,) or (n_stencil, *out_shape) in 7dbea2047630bc11e1e024973bf9596677d97db5

src/derivkit/tabulated_model/one_d.py
nikosarcevic: great idea, the shape logic is easier to read as a match with a single return. I refactored parse_xy_table to use match (n_rows, n_cols) and return (x, y) at the end in e0e616544597fe3b273f8017b0ea3e708587b312

src/derivkit/tabulated_model/one_d.py
nikosarcevic: I kept parse_xy_table as a pure (x, y) parser and tabulated1d_from_table as the small convenience wrapper that also handles extrpaolate/fill_value. The 2D restriction is intentional for simple csv txt-style tables; higher-rank data should be reshaped and passed to Tabulated1DModel directly. Docstrings updated in 770ebaa02aee384055388845317a34b985774c4e

tests/test_tabulated_one_d_calculus.py
nikosarcevic: fixed even more (removed underscore) in 96c216da92b64f73d75ba950f2f2ccaa566946a1

tests/test_tabulated_one_d_forecasting.py
nikosarcevic: also removed _ in make fucntons 918e3d03939f70371486d0f5232093c0b799753e

src/derivkit/finite/core.py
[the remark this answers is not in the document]
lonbar: Thanks for clarifying :)

src/derivkit/tabulated_model/one_d.py
```diff
@@ -0,0 +1,236 @@
...
+        extrapolate: bool = False,
+        fill_value: float | None = None,
+    ) -> None:
+        """Initializes a tabulated 1D interpolation model.
+
+        Args:
+            x: Strictly increasing tabulated x values with shape ``(N,)``.
+            y: Tabulated y values with shape ``(N,)`` (scalar) or ``(N, ...)`` (vector/tensor).
+               The first dimension must match ``x``.
+            extrapolate: Whether to allow evaluation outside the x range. Default is False.
```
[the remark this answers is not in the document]
lonbar: I think that's okay. I guess you'd want this in case the region outside of the range is non-physical.

src/derivkit/calculus/hessian.py
```diff
@@ -390,7 +392,14 @@ def _mixed_partial_value(
...
     kit1 = DerivativeKit(partial_vec1, float(theta[i]))
-    return float(kit1.differentiate(order=1, method=method, n_workers=n_workers, **dk_kwargs))
+    val = kit1.differentiate(
+        order=1,
+        method=method,
+        n_workers=n_workers,
+        **dk_kwargs,
+    )
+    val_arr = np.asarray(val, dtype=float)
+    return float(val_arr.ravel(order="C")[0])
```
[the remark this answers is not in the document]
lonbar: Do you want to propagate your fix from https://github.com/derivkit/derivkit/commit/6c2737beaa798bd015a61b22335fc05302846897 to here?

src/derivkit/calculus/hessian.py
```diff
@@ -390,7 +391,13 @@ def _mixed_partial_value(
...
     partial_vec1 = get_partial_function(function, i, theta)
     kit1 = DerivativeKit(partial_vec1, float(theta[i]))
-    return float(kit1.differentiate(order=1, method=method, n_workers=n_workers, **dk_kwargs))
+    val = kit1.differentiate(
+        order=1,
+        method=method,
+        n_workers=n_workers,
+        **dk_kwargs,
+    )
+    return float(np.asarray(val, dtype=float).item())
```
lonbar:
```suggestion
    return float(val)
```

src/derivkit/calculus/hessian.py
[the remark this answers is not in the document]
lonbar: Done is 9a46f52f41b0d1c6d53e21dc98013d98bfac6862.

lonbar (approved):

PR #261, opened 2025-12-02
Add the tabulated model methods to the docs index

lonbar:
This is just a small addition that mentions the data interpolator on the front page, without much detail. More details can be found in the API docs. We may want to add more in the guides section.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -77,6 +77,7 @@ DerivKit is organized into four layers:
    * Tabulated 1D interpolator
    * Tabulated 1D from table
 
+   Supports scalar, vector or tensor outputs for all derivatives.
    Data can be read from memory or from a text file.
 
 Installation


docs/index.rst
```diff
@@ -70,6 +70,14 @@ DerivKit is organized into four layers:
...
    Handles flattening, reshaping, and validation.
 
+.. dropdown:: **5. Tabulate layer**
+
+   Formats and interpolates between discrete data points:
+
+   * Tabulated 1D interpolator
+   * Tabulated 1D from table
+
+   Data can be read from memory or from a text file.
```
nikosarcevic: I would add:

Supports scalar/vector/tensor outputs and full derivative support.
 Data can be read from memory or from a text file.

nikosarcevic (approved):

Issue #195, opened 2025-10-29
Split expansions so each product has its own script


nikosarcevic:
Currently, all products (fisher bias, dali, fisher) are living in one file: expansions.py.
Since this file is now very long and hard to follow, it would be good to move functions into their own modules within the forecasting dir.
This will be good for future extensions since we only need ot add one file and work in it instead of jsut adding to this one gigantic expansions file.

This is not urgent.

Also make sure to also split the unit tests into its own files.
Naming convention probably: test_kitname_methodname.py
aka tests_forecasting_fisher_bias.py something like this.

PR #262, opened 2025-12-02
Expansions restructure

ctrendafilova:
Restructuring `expansions.py` to address #195 .

I have separated the two Fisher bias functions -- `build_fisher_bias` and `build_delta_nu` -- into a separate file, `fisher_bias.py`.

This is my first attempt, so please review and give feedback on how to improve the changes.

Also, I didn't know how to correctly specify the `LikelihoodExpansion` type in lines 20 and 137 of `fisher_bias.py`.

### DIFF
--- a/src/derivkit/forecasting/expansions.py
+++ b/src/derivkit/forecasting/expansions.py
@@ -17,7 +17,7 @@ from numpy.typing import ArrayLike, NDArray
 
 from derivkit.calculus_kit import CalculusKit
 from derivkit.utils.linalg import invert_covariance, solve_or_pinv
-from derivkit.forecasting import fisher_bias
+from derivkit.forecasting import fisher, dali
 
 
 class LikelihoodExpansion:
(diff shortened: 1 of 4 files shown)


src/derivkit/forecasting/expansions.py
```diff
@@ -255,29 +241,11 @@ def _build_dali(
             d2: NDArray[np.float64],
             invcov: NDArray[np.float64],
     ) -> tuple[NDArray[np.float64], NDArray[np.float64]]:
-        """Assemble the doublet-DALI tensors (G, H) from first- and second-order derivatives.
-
-        Computes:
-            G_abc = Σ_{i,j} d2[a,b,i] · invcov[i,j] · d1[c,j]
-            H_abcd = Σ_{i,j} d2[a,b,i] · invcov[i,j] · d2[c,d,j]
+        """Return the doublet-DALI tensors (G, H).
```
nikosarcevic: I think here we need to use "Returns" instead of "Return" as this is a standard we imposed.

src/derivkit/forecasting/fisher.py
```diff
@@ -16,6 +16,25 @@
 from derivkit.utils.linalg import solve_or_pinv
 
 
+def _build_fisher(expansion, d1, invcov):
```
nikosarcevic: same here -- Assembles

PR #271, opened 2025-12-06
Split expansions (step 2): Create forecast core utilities

nikosarcevic:
Extract low-level Fisher/DALI helpers into forecast_core.py.
No public API changes; LikelihoodExpansion behavior is unchanged.

This builds on top of step1. So after step1 is merged, I cna rebase and only 2 files are actually changed here.


lonbar:
This PR is no longer necessary since its changes were added by #274.

PR #272, opened 2025-12-06
Update version dynamically

lonbar:
The PR adds dynamic versioning. The simplest way of creating new versions is by using the semantic versioning scheme prepended by a `v` as described in [the setuptools-scm basic usage](https://setuptools-scm.readthedocs.io/en/latest/):
```
git tag v0.1.0
pip install .
python
>>> import derivkit
>>> derivkit.__version__
'0.1.0'
```
Users can still create installations at arbitrary commits, which results in a fudged version number. For example, moving to commit 0188ce99 before running the `pip install` command results in a version number of `derivkit-0.1.dev854+g0188ce99f`. This should make it a bit easier to keep track of different version numbers.

I have also added support for different versions of the documentation. Currently each branch and each git tag will result in a separate copy of the documentation, which can be accessed through the sidebar.


nikosarcevic (approved):

lonbar:
Probably we will eventually want to limit the number of versions (say, each tag and the main branch?).

Also, we will need to add a redirect page to the `gh-pages` branch. It otherwise isn't clear what the version to be desplayedshould be.

PR #283, opened 2025-12-08
Split expansions (step 9): Replaced calls to LikelihoodExpansions to calls from dali.py in test_forecast_kit_dali

nikosarcevic:
This PR builds on top of Step 8 #282 

In this PR I have switched from using LikelihoodExpansions dali to dali.py code in test_forecast_kit_dali.py

PR #284, opened 2025-12-08
Split expansions (step 10): Replaced calls to LikelihoodExpansions to calls fom forecast_core in forecastkit_parallel test

nikosarcevic:
This PR builds on top of PR Step 9  #283 

In this PR I have changed the calls from LikelihoodExpansion to calls from forecast_core to update the test_forecast_kit_parallel

### DIFF
--- a/tests/test_forecast_kit_parallel.py
+++ b/tests/test_forecast_kit_parallel.py
@@ -257,7 +257,7 @@ def test_parallel_is_deterministic_across_runs_order2(extra_threads_ok):
 
 
 def test_parallel_preserves_order_with_prime_length_order2(extra_threads_ok):
-    """Verifys derivative ordering is preserved even with uneven parallel splits (order=2).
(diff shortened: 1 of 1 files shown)


src/derivkit/forecasting/forecast_core.py
```diff
@@ -24,6 +24,7 @@
 
 __all__ = [
     "get_forecast_tensors",
+    "_get_derivatives",
```
lonbar: This is a private function, so shouldn't we write tests around it?

lonbar: I remember that we discussed this before. I don't quite remember what the reasoning was that we've left it as it is. I'Also, I'm not sure if we want to expose a private function like this.

lonbar:
```suggestion
```

lonbar (approved):

PR #285, opened 2025-12-08
Split expansions (step 11):  Removed LikelihoodExpansions and any mention of it

nikosarcevic:
This PR builds on top of step 10 #284 

Removes the deprecated expansions.py module and eliminates all remaining references to LikelihoodExpansion. Updates docstrings and wrapper descriptions to reference the new forecast-core utilities.


lonbar (approved):
Looks good to me. Again, I've removed the commits shared with the previous steps. This is the one remaining commit, so I think it's fine. But do double check.

lonbar (approved):

PR #292, opened 2025-12-10
Docs: remove unimplemented Sellentin–Heavens likelihood (#281)

### LINKED ISSUE #281: fix sellentin likelihood in the docs
currently likeihood in the docs for sellentin is named wrongly (no alaln heavens) and also the correct equation from the paper needs to be included.

nikosarcevic:
This PR removes the Sellentin–Heavens likelihood section from the LikelihoodKit documentation.
The method is not implemented in DerivKit, and the docs were misleading by implying support.

This closes #281

### DIFF
--- a/docs/guide/likelihoods.rst
+++ b/docs/guide/likelihoods.rst
@@ -103,5 +103,4 @@ calculations:
 - likelihoods can be swapped without changing forecasting code.
 
 This modularity allows users to begin with simple Gaussian forecasts
-and later transition to Poisson or Sellentin–Heavens likelihoods
-without rewriting their models.
+and later transition to Poisson likelihood without rewriting their models.


lonbar:
I don't think that this solves #281. The docs state that this likelihood is planned. Also, I think we discussed that we wanted to rename how we refer to it in the docs.

nikosarcevic:
> I don't think that this solves #281. The docs state that this likelihood is planned. Also, I think we discussed that we wanted to rename how we refer to it in the docs.

yeah i mean i should update the issue. you said not to add any empty files in the likelihood so we do have a to do list and now we can remove this until it is implelented. ill change the issue wording to be strict.

ETA: fixed the wording in the issue.

PR #295, opened 2025-12-11
Expose tabulated (x, y) data through DerivativeKit

### LINKED ISSUE #294: Expose Tabulated derivs into API properly
Currently, we have Tabulated derivatives live in their own little module that is not directly exposed in the API.

The idea is to somehow expose Tabulated thru DerivativeKit. 

Things to keep in mind are:

- keep it as simple as possible (no abstractions)
- figure out how to make DerivativeKit init smart so it "knows" if you are passing the function or tabulated data (or figure out the kwargs init)
- circular imports: currently Tabualted calls DerivativeKit so if we call it inside DerivativeKit we have a circular import

nikosarcevic:
This PR makes DerivativeKit aware of tabulated models so users can do:
```
dk = DerivativeKit(x0=0.5, tab_x=x_tab, tab_y=y_tab)
dk.differentiate(method="finite", order=1)
```
What’s included

- Added tab_x / tab_y kwargs to DerivativeKit.
- Auto-wrap tabulated data into Tabulated1DModel internally (no need to import it explicitly).
- Preserved the “callable function” API with strict mutual-exclusion (function vs tabulated inputs).
- Updated tests for both the tabulated model and DerivativeKit dispatch to cover this path.
- No API breaks

This keeps the surface API simple while avoiding circular imports and keeps Tabulated1DModel self-contained.

This closes #294

### DIFF
--- a/src/derivkit/tabulated_model/one_d.py
+++ b/src/derivkit/tabulated_model/one_d.py
@@ -26,9 +26,6 @@ class Tabulated1DModel:
     """1D interpolator for tabulated data.
 
     Interpolates a tabulated function ``y(x)`` using ``numpy.interp``.
-    To compute derivatives of the interpolated function, use
-    :class:`DerivativeKit` with ``function=model`` or the model instance
-    as the callable.
 
     Here ``x`` is a one-dimensional grid of length ``N``, and the first
     dimension of ``y`` must also have length ``N``. All remaining


src/derivkit/derivative_kit.py
```diff
@@ -18,6 +18,17 @@
...
+    Using tabulated data directly:
+
+        >>> import numpy as np
+        >>> from derivkit.derivative_kit import DerivativeKit
+        >>>
+        >>> x_tab = np.array([0.0, 1.0, 2.0, 3.0])
+        >>> y_tab = x_tab**2
+        >>> dk = DerivativeKit(x0=0.5, tab_x=x_tab, tab_y=y_tab)
+        >>> # First derivative via the finite difference method:
+        >>> # dk.differentiate(order=1, method="finite", extrapolation="ridders")  # doctest: +SKIP
```
lonbar:
```suggestion
        >>> # dk.differentiate(order=1, method="finite", extrapolation="ridders")
```

src/derivkit/tabulated_model/one_d.py
```diff
@@ -9,17 +9,17 @@
 * Direct construction with ``(x, y)`` arrays of shape ``(N,)`` and ``(N, ...)``.
 * :func:`tabulated1d_from_table` for simple 2D tables containing x and one
   or more y components in columns.
+
+To differentiate interpolated models, pass the model instance as
+``function`` to :class:`DerivativeKit`.
```
lonbar:
```suggestion
```

src/derivkit/tabulated_model/one_d.py
```diff
@@ -29,6 +29,9 @@ class Tabulated1DModel:
     """1D interpolator for tabulated data.
 
     Interpolates a tabulated function ``y(x)`` using ``numpy.interp``.
+    To compute derivatives of the interpolated function, use
+    :class:`DerivativeKit` with ``function=model`` or the model instance
+    as the callable.
```
lonbar:
```suggestion
```

PR #305, opened 2025-12-16
GetDist Integration PR1 - Add forecasting priors and numeric helpers

nikosarcevic:
This PR adds a small forecasting priors module with unit tests, plus the minimal numeric helpers it depends on.
It is intentionally scoped to priors only; expansions extension in forecasting and sampling (getdist integration) will come in the next batches

### DIFF
--- a/docs/derivkit.forecasting.priors.core.rst
+++ b/docs/derivkit.forecasting.priors.core.rst
@@ -1,0 +1,7 @@
+derivkit.forecasting.priors.core module
+=======================================
+
+.. automodule:: derivkit.forecasting.priors.core
+   :members:
+   :show-inheritance:
+   :undoc-members:
--- a/docs/derivkit.forecasting.priors.rst
+++ b/docs/derivkit.forecasting.priors.rst
@@ -1,0 +1,18 @@
+derivkit.forecasting.priors package
+===================================
+
+Submodules
+----------
+
+.. toctree::
+   :maxdepth: 4
+
+   derivkit.forecasting.priors.core
+
+Module contents
+---------------
+
+.. automodule:: derivkit.forecasting.priors
+   :members:
+   :show-inheritance:
+   :undoc-members:
--- a/docs/derivkit.forecasting.rst
+++ b/docs/derivkit.forecasting.rst
@@ -1,6 +1,14 @@
 derivkit.forecasting package
 ============================
 
+Subpackages
+-----------
+
+.. toctree::
+   :maxdepth: 4
+
+   derivkit.forecasting.priors
+
 Submodules
 ----------
 


src/derivkit/forecasting/priors/__init__.py
lonbar: I'm wondering if this is going a bit overboard with the subpackages.

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+def evaluate_logprior(
+    theta: ArrayLike,
+    logprior: Callable[[NDArray[np.floating]], float] | None,
+) -> float:
+    """Evaluates a log-prior with robust handling of invalid values.
+
+    If no prior is provided, this function represents an improper flat prior
+    and returns zero. If a prior is provided, its value is interpreted as a
+    log-density defined up to an additive constant. Any non-finite value
+    (e.g., ``-inf`` or ``nan``) is treated as zero prior probability.
```
lonbar:
```suggestion
    (e.g., ``-np.inf`` or ``np.nan``) is treated as zero prior probability.
```

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+        The log-prior value at ``theta``, or ``-inf`` if the prior assigns
+        zero probability to that point.
+    """
+    if logprior is None:
+        return 0.0
+    v = float(logprior(np.asarray(theta)))
+    return v if np.isfinite(v) else -np.inf
+
+
+def in_bounds_checker(
```
lonbar:
```suggestion
def is_in_bounds(
```
I feel this is slightly more ideomatic, as in this case you'd have a test like
```
if not is_in_bounds:
   # do something
```

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    """
+    if bounds is None:
+        return True
+    if len(bounds) != theta.size:
+        raise ValueError(f"bounds length {len(bounds)} != theta length {theta.size}")
+
+    for t, (lo, hi) in zip(theta, bounds):
+        if (lo is not None and t < lo) or (hi is not None and t > hi):
+            return False
+    return True
```
lonbar: This isn't Pythonic. A quick example using list comprehension:
```suggestion
    result = True
    if bounds is not None:
        if len(bounds) != theta.size:
            raise ValueError(f"bounds length {len(bounds)} != theta length {theta.size}")

        is_within_bounds = [True for x in theta for y in bounds if y is not None and x > y[0] and x < y[1]]
        all_in_bounds = not False in is_within_bounds
        
    return all_in_bounds
```
If you don't want to use list comprehension you could rewrite the for loop like
```
    for t, (lo, hi) in zip(theta, bounds):
        if (lo is not None and t < lo) or (hi is not None and t > hi):
            result = False
            break
```
This limits the branching.

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    """Compute log(sum(exp(x))) stably for a 1D array.
+
+    This is a minimal, dependency-free alternative to scipy.special.logsumexp
+    for the common 1D case (e.g., mixture models in log-space).
+
+    Args:
+        x: 1D array-like values.
+
+    Returns:
+        Stable logsumexp as float.
```
lonbar: Please clarify "stable"

lonbar: I think it's better to avoid such qualifiers.

src/derivkit/utils/numerics.py
lonbar: I stopped suggesting changes for this at some point, but please be consistent in using `float` or `np.floating`.

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    *,
+    term: Callable[[NDArray[np.floating]], float],
+    bounds: Sequence[tuple[float | None, float | None]] | None,
+) -> float:
+    """Evaluates a log-density term with hard-support truncation.
+
+    This helper enforces a top-hat support region defined by ``bounds``. If
+    ``theta`` lies outside the allowed region, the result is ``-inf`` to denote
+    zero probability. If ``theta`` lies inside the region, the provided term is
+    evaluated and any non-finite output is treated as ``-inf``.
```
lonbar:
```suggestion
    This helper enforces a top-hat support region defined by ``bounds``. If
    ``theta`` lies outside the allowed region, the result is ``-np.inf`` to denote
    zero probability. If ``theta`` lies inside the region, the provided term is
    evaluated and any non-finite output is treated as ``-np.inf``.
```

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+
+    Returns:
+        The term value at ``theta`` if inside bounds and finite; otherwise
+        ``-inf``.
+    """
+    th = np.asarray(theta, dtype=float)
+    if not in_bounds_checker(th, bounds):
+        return -np.inf
+    v = float(term(th))
+    return v if np.isfinite(v) else -np.inf
```
lonbar:
```suggestion
    v = -np.inf
    if in_bounds_checker(th, bounds):
        v = float(term(th))

    return v if np.isfinite(v) else -np.inf
```

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+
+    Returns:
+        Sum of term values at ``theta`` if all are finite; otherwise ``-inf``.
+    """
+    th = np.asarray(theta, dtype=float)
+    total = 0.0
+    for f in terms:
+        v = float(f(th))
+        if not np.isfinite(v):
+            return -np.inf
```
lonbar:
```suggestion
        if not np.isfinite(v):
            total = np.inf
            break
```

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    if len(terms) == 0:
+        raise ValueError("sum_terms requires at least one term")
+    return partial(_sum_terms_impl, terms=terms)
+
+
+def as_1d_float_array(x: ArrayLike, *, name: str = "x") -> NDArray[np.float64]:
+    """Convert input to a 1D float array.
+
+    This is a small convenience for model/likelihood/prior code that expects
+    parameter vectors. It performs a minimal shape check (must be 1D) and
```
lonbar:
```suggestion
    parameter vectors. It performs a shape check (the array must be 1D) and
```

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+        ValueError: If the converted array is not 1D.
+    """
+    arr = np.asarray(x, dtype=float)
+    if arr.ndim != 1:
+        raise ValueError(f"{name} must be 1D, got shape {arr.shape}")
+    return arr.astype(np.float64, copy=False)
+
+
+def get_index_value(theta: ArrayLike, index: int, *, name: str = "theta") -> float:
+    """Return theta[index] as a float with a clear bounds error.
```
lonbar:
```suggestion
    """Returns the value of an array specified by the given index.
```

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+        return float(-0.5 * z * z - lx)
+
+    if kind == "beta":
+        alpha = float(a)
+        beta = float(b)
+        if alpha <= 0.0 or beta <= 0.0:
+            raise ValueError("alpha and beta must be > 0")
+        return float((alpha - 1.0) * np.log(x) + (beta - 1.0) * np.log1p(-x))
+
+    raise ValueError(f"unknown prior kind '{kind}'")
```
lonbar: Please use a `match`. It would be preferable to have a single return statement.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+
+def prior_none() -> Callable[[NDArray[np.floating]], float]:
+    """Constructs an improper flat prior (constant log-density).
+
+    This prior has density proportional to 1 everywhere.
+
+    Returns:
+        Callable log-prior: logp(theta) -> 0.0
+    """
+    return _prior_none_impl
```
lonbar: This function is only used once. Since it is this simple it may as well be removed.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+        ValueError: If `sigma_log` is not positive.
+    """
+    s = float(sigma_log)
+    if s <= 0.0:
+        raise ValueError("sigma_log must be > 0")
+    return partial(_prior_1d_impl, index=int(index), domain="positive", kind="log_normal", a=float(mean_log), b=s)
+
+
+def prior_beta(*, index: int, alpha: float, beta: float) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a Beta prior for a single parameter in (0, 1).
```
lonbar: Please clarify "Beta prior"

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+
+    This prior has density proportional to x^(alpha-1) * (1-x)^(beta-1) for x in (0, 1).
```
lonbar: If this is the "Beta prior" please make this explicit. Why is it proportional but not equal?

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+
+    Args:
+        index: Index of the parameter to which the prior applies.
+        alpha: Alpha shape parameter (> 0).
+        beta: Beta shape parameter (> 0).
```
lonbar:
```suggestion
        alpha: Alpha shape parameter (must be greater than 0).
        beta: Beta shape parameter (must be greater than 0).
```

src/derivkit/forecasting/priors/core.py
lonbar: Also, consider using either code formatting (backticks) or math formatting

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    """Internal helper that implements an improper flat prior.
+
+    Args:
+        theta: Parameter vector (unused).
+
+    Returns:
+        0.0
+    """
+    _ = theta  # unused
+    return 0.0
```
lonbar:
```suggestion
    return 0.0
```

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+
+
+def _prior_none_impl(theta: NDArray[np.floating]) -> float:
+    """Internal helper that implements an improper flat prior.
+
+    Args:
+        theta: Parameter vector (unused).
+
+    Returns:
+        0.0
```
lonbar: Please clarify this.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    "prior_gaussian_mixture",
+    "build_prior"
+]
+
+
+def _prior_none_impl(theta: NDArray[np.floating]) -> float:
+    """Internal helper that implements an improper flat prior.
+
+    Args:
+        theta: Parameter vector (unused).
```
lonbar: If it is unused it may as well be removed.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    "prior_half_normal",
+    "prior_half_cauchy",
+    "prior_log_normal",
+    "prior_beta",
+    "prior_gaussian_mixture",
+    "build_prior"
+]
+
+
+def _prior_none_impl(theta: NDArray[np.floating]) -> float:
```
lonbar: This function isn't used much, so why not remove it.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    elif domain == "nonnegative":
+        if x < 0.0:
+            return -np.inf
+    elif domain == "unit_open":
+        if x <= 0.0 or x >= 1.0:
+            return -np.inf
+    else:
+        raise ValueError(f"unknown domain '{domain}'")
+
+    # We define different prior shapes here
```
lonbar: Redundant comment.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    if kind == "log_uniform":
+        return float(-np.log(x))
+
+    if kind == "half_normal":
+        sigma = float(a)
+        # We need to validate here too, in case someone partials directly to _prior_1d_impl
```
lonbar: Please clarify this.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    return _prior_none_impl
+
+
+def prior_uniform(
+    *,
+    bounds: Sequence[tuple[float | None, float | None]],
+) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a uniform prior with hard bounds.
+
+    This prior has density proportional to 1 within the specified bounds
```
lonbar:
```suggestion
    This prior has a density proportional to 1 within the specified bounds
```

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    """
+    return partial(_prior_1d_impl, index=int(index), domain="positive", kind="log_uniform")
+
+
+def prior_jeffreys(*, index: int) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a Jeffreys prior for a single positive scale parameter.
+
+    The Jeffreys prior is defined using the Fisher information of the parameter.
+    For a positive scale parameter, this leads to a prior proportional to 1/x
+    (with x being the scale parameter that is > 0), which is the same functional
```
lonbar: please use full sentences and correct english

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+
+    Note:
+        Although the implementation matches ``prior_log_uniform``, the name
+        ``prior_jeffreys`` emphasizes the motivation (reparameterization
+        invariance for scale parameters) rather than “uniform in log-space”.
+    """
+    return partial(_prior_1d_impl, index=int(index), domain="positive", kind="log_uniform")
+
+
+def prior_half_normal(*, index: int, sigma: float) -> Callable[[NDArray[np.floating]], float]:
```
lonbar: I'm a bit confused about what the usecase of this is. The user could simply use a normal distribution, just with non-negative parameter values.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    """Constructs a half-normal prior for a single non-negative parameter.
+
+    This prior has density proportional to exp(-0.5 * (x/sigma)^2) for x >= 0.
```
lonbar: Please clarify what the proportionality constant is, why it is neglected, etc.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    Raises:
+        ValueError: If `scale` is not positive.
+    """
+    s = float(scale)
+    if s <= 0.0:
+        raise ValueError("scale must be > 0")
+    return partial(_prior_1d_impl, index=int(index), domain="nonnegative", kind="half_cauchy", a=s)
+
+
+def prior_log_normal(
```
lonbar: You could use the normal distribution from likelihoods.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    Args:
+        spec: Prior specification dictionary.
+
+    Returns:
+        Callable log-prior term: logp(theta) -> float.
+
+    Notes:
+        - For name="uniform", bounds must be provided via params={"bounds": ...}.
+          The top-level "bounds" key is reserved for optional *extra* hard bounds
+          on non-uniform terms.
```
lonbar: I'm not sure why this needs a `Notes` header. You could mention all of this directly under the summary line.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    """Builds a single prior term from a configuration dictionary.
+
+    Expected format:
+        {"name": "<prior_name>", "params": {...}, "bounds": optional_bounds}
+
+    Args:
+        spec: Prior specification dictionary.
+
+    Returns:
+        Callable log-prior term: logp(theta) -> float.
```
lonbar: Please elaborate a bit more on what this is, using full sentences and correct English.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    if name == "uniform":
+        # allow either params["bounds"] or spec["bounds"] (but not both)
+        pb = params.get("bounds", None)
+        if pb is not None and term_bounds is not None:
+            raise ValueError(
+                "uniform prior: provide bounds via either params['bounds'] or top-level 'bounds', not both")
+        if pb is None and term_bounds is None:
+            raise ValueError("uniform prior requires bounds")
+        b = pb if pb is not None else term_bounds
+        return prior_uniform(bounds=b)
```
lonbar: Please instead define a value here which can be returned at the end of the function.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+          * ("prior_name", {params})
+          * {"name": "prior_name", "params": {...}, "bounds": optional_term_bounds}
+      - `bounds`: optional global hard bounds applied to the *combined* prior
+
+    Conventions:
+      - If `terms` is None or empty:
+          * if `bounds` is None -> improper flat prior (prior_none)
+          * else -> uniform top-hat prior over `bounds` (prior_uniform)
+      - "uniform" priors must be passed as ("uniform", {"bounds": ...}) or the dict equivalent.
+        Global `bounds` is still allowed (it just adds another hard gate).
```
lonbar: Consider placing this directly below the summary line. There doesn't need to be a separate `Conventions` field.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+          ],
+          bounds=[(0.0, None), (None, None), (0.0, 1.0)],
+      )
+    """
+    term_list = [] if terms is None else list(terms)
+
+    # Empty -> either flat or bounded-uniform
+    if len(term_list) == 0:
+        base = prior_none() if bounds is None else prior_uniform(bounds=bounds)
+        return base
```
lonbar: Please assign a variable here and return it at the end of the function.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+        name, params = t
+        if not isinstance(params, dict):
+            raise TypeError("Term params must be a dict.")
+        specs.append({"name": str(name), "params": dict(params)})
+
+    # Use existing builder for dict-spec terms (supports per-term bounds, uniform special case)
+    built_terms = [_make_prior_term(s) for s in specs]
+    combined = sum_terms(*built_terms)
+
+    # Global bounds apply last
```
lonbar: This comment seems redundant. Please elaborate a bit if more context is needed.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+        if not isinstance(t, (tuple, list)) or len(t) != 2:
+            raise TypeError(
+                "Each term must be either a dict spec or a (name, params) tuple/list of length 2."
+            )
+        name, params = t
+        if not isinstance(params, dict):
+            raise TypeError("Term params must be a dict.")
+        specs.append({"name": str(name), "params": dict(params)})
+
+    # Use existing builder for dict-spec terms (supports per-term bounds, uniform special case)
```
lonbar: I'm not sure I understand this. What is the existing builder?

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    if s <= 0.0:
+        raise ValueError("sigma must be > 0")
+    return partial(_prior_1d_impl, index=int(index), domain="nonnegative", kind="half_normal", a=s)
+
+
+def prior_half_cauchy(*, index: int, scale: float) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a half-Cauchy prior for a single non-negative parameter.
+
+    This prior has density proportional to 1 / (1 + (x/scale)^2) for x >= 0,
+    with x being the parameter.
```
lonbar: Please consider using math or code mode.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+
+def prior_gaussian_diag(
+    *,
+    mean: NDArray[np.floating],
+    sigma: NDArray[np.floating],
+) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a diagonal Gaussian prior (up to an additive constant).
+
+    This prior has density proportional to
+    exp(-0.5 * sum_i ((x_i - mean_i) / sigma_i)^2), with independent components.
```
lonbar: Please consider using math or code mode.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+def prior_gaussian(
+    *,
+    mean: NDArray[np.floating],
+    cov: NDArray[np.floating] | None = None,
+    inv_cov: NDArray[np.floating] | None = None,
+) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a multivariate Gaussian prior (up to an additive constant).
+
+    This prior has density proportional to
+    exp(-0.5 * (x - mean)^T @ inv_cov @ (x - mean)) with inv_cov being the
```
lonbar: Please consider using math or code mode.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+
+    logprior(theta) -> float
+
+The return value is interpreted as a log-density defined up to an additive
+constant. Returning ``-np.inf`` denotes zero probability (hard exclusion).
+
+These priors are designed to be used when constructing log-posteriors for
+sampling (e.g., Fisher/DALI approximate posteriors) or when evaluating posterior
+surfaces. GetDist plots what is provided; priors must be applied when generating
+samples/gaussians (or explicitly added to the Fisher), not automatically inferred.
```
ctrendafilova: I find the last sentence a bit confusing to parse. What do you mean by "generating gaussians" in this context?

I'm also not sure what is meant by the parenthetical about the Fisher matrix.

src/derivkit/forecasting/priors/__init__.py
nikosarcevic: Good point. The intent is to keep the public surface small and avoid circular imports by grouping related prior builders and utilities, not to create a deep hierarchy. I’m happy to flatten this (e.g., keep priors as a single module or reduce subpackages) if that’s preferred in derivkit.
My only suggestion would be to do this after the current PR series is merged, to avoid unnecessary churn across multiple dependent PRs.

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
     b = np.asarray(b, dtype=float)
     denom = np.maximum(1.0, np.maximum(np.abs(a), np.abs(b)))
     return float(np.max(np.abs(a - b) / denom))
+
+
+def evaluate_logprior(
+    theta: ArrayLike,
+    logprior: Callable[[NDArray[np.floating]], float] | None,
+) -> float:
+    """Evaluates a log-prior with robust handling of invalid values.
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
     b = np.asarray(b, dtype=float)
     denom = np.maximum(1.0, np.maximum(np.abs(a), np.abs(b)))
     return float(np.max(np.abs(a - b) / denom))
+
+
+def evaluate_logprior(
+    theta: ArrayLike,
+    logprior: Callable[[NDArray[np.floating]], float] | None,
+) -> float:
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

and 
867d9f64b592cae8aced383647e18c77e0fafa12

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+        logprior: Callable returning the log-prior density, or ``None`` to
+            indicate an uninformative (flat) prior.
+
+    Returns:
+        The log-prior value at ``theta``, or ``-inf`` if the prior assigns
+        zero probability to that point.
+    """
+    if logprior is None:
+        return 0.0
+    v = float(logprior(np.asarray(theta)))
```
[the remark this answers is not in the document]
nikosarcevic: https://github.com/derivkit/derivkit/commit/243daefdf750f13484c4413d14c3786d003cf142

and
https://github.com/derivkit/derivkit/commit/867d9f64b592cae8aced383647e18c77e0fafa12

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
     return float(np.max(np.abs(a - b) / denom))
+
+
+def evaluate_logprior(
+    theta: ArrayLike,
+    logprior: Callable[[NDArray[np.floating]], float] | None,
+) -> float:
+    """Evaluates a log-prior with robust handling of invalid values.
+
+    If no prior is provided, this function represents an improper flat prior
```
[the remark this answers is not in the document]
nikosarcevic: 2cb353301969233717ce98c358da35b00c4ed090

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
     b = np.asarray(b, dtype=float)
     denom = np.maximum(1.0, np.maximum(np.abs(a), np.abs(b)))
     return float(np.max(np.abs(a - b) / denom))
+
+
+def evaluate_logprior(
```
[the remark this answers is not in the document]
nikosarcevic: Users can write priors directly; this helper exists to centralize the None so ti returns flat prior convention and the non-finite then returns -np.inf mapping, so callers don’t duplicate that boilerplate and we keep consistent semantics everywhere.

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    """
+    if logprior is None:
+        return 0.0
+    v = float(logprior(np.asarray(theta)))
+    return v if np.isfinite(v) else -np.inf
+
+
+def in_bounds_checker(
+    theta: NDArray[np.floating],
+    bounds: Sequence[tuple[float | None, float | None]] | None,
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+) -> bool:
+    """Tests whether parameters lie within optional per-parameter bounds.
```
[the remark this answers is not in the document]
nikosarcevic: The check is written this way so callers don’t need to special-case the absence of bounds. By convention, bounds=None means “no constraints”, and is_in_bounds then returns True. This allows patterns like if is_in_bounds(theta, bounds): ... to work uniformly whether bounds are provided or not, and keeps the logic centralized in one place.

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    all points are considered valid.
+
+    Args:
+        theta: Parameter vector to test.
+        bounds: Optional sequence of ``(lower, upper)`` pairs, one per parameter.
+            Use ``None`` for an unconstrained lower or upper limit.
+
+    Returns:
+        True if all parameters satisfy their bounds (or if ``bounds`` is ``None``);
+        otherwise False.
```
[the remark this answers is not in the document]
nikosarcevic: 2cb353301969233717ce98c358da35b00c4ed090 it is now changed

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    if j < 0 or j >= th.size:
+        raise IndexError(f"{name} index {j} out of bounds for length {th.size}")
+    return float(th[j])
+
+
+def logsumexp_1d(x: ArrayLike) -> float:
+    """Compute log(sum(exp(x))) stably for a 1D array.
+
+    This is a minimal, dependency-free alternative to scipy.special.logsumexp
+    for the common 1D case (e.g., mixture models in log-space).
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    return True
+
+
+def _apply_hard_bounds_impl(
+    theta: NDArray[np.floating],
+    *,
+    term: Callable[[NDArray[np.floating]], float],
+    bounds: Sequence[tuple[float | None, float | None]] | None,
+) -> float:
+    """Evaluates a log-density term with hard-support truncation.
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
nikosarcevic: explained 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+
+    Args:
+        theta: Parameter vector at which to evaluate the term.
+        term: Callable returning a log-density contribution.
+        bounds: Optional sequence of ``(lower, upper)`` pairs defining the
+            allowed support region.
+
+    Returns:
+        The term value at ``theta`` if inside bounds and finite; otherwise
+        ``-inf``.
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    term: Callable[[NDArray[np.floating]], float],
+    *,
+    bounds: Sequence[tuple[float | None, float | None]] | None = None,
+) -> Callable[[NDArray[np.floating]], float]:
+    """Returns a bounded version of a log-density term.
+
+    If ``bounds`` is provided, the returned callable represents the same log
+    term but with compact support: points outside the bounds are assigned
+    ``-inf`` (zero probability). If ``bounds`` is ``None``, the original term
+    is returned unchanged.
```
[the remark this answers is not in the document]
nikosarcevic: This helper is meant to be composable. Callers can pass bounds=None when they want to apply hard bounds conditionally or reuse the same code path regardless of whether bounds are present. In that case, returning the original term unchanged avoids special-casing in the caller.

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+def _sum_terms_impl(
+    theta: NDArray[np.floating],
+    *,
+    terms: tuple[Callable[[NDArray[np.floating]], float], ...],
+) -> float:
+    """Evaluates a sum of log-density terms.
+
+    This helper is meant for constructing composite priors/likelihood pieces as
+    additive log terms. If any term evaluates to a non-finite value, the sum is
+    treated as ``-inf`` to indicate that the combined density assigns zero
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+
+def _sum_terms_impl(
+    theta: NDArray[np.floating],
+    *,
+    terms: tuple[Callable[[NDArray[np.floating]], float], ...],
+) -> float:
+    """Evaluates a sum of log-density terms.
+
+    This helper is meant for constructing composite priors/likelihood pieces as
+    additive log terms. If any term evaluates to a non-finite value, the sum is
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+        if not np.isfinite(v):
+            return -np.inf
+        total += v
+    return total
+
+
+def sum_terms(
+    *terms: Callable[[NDArray[np.floating]], float],
+) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a composite log term by summing multiple contributions.
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+            Use ``None`` for an unconstrained lower or upper limit.
+
+    Returns:
+        True if all parameters satisfy their bounds (or if ``bounds`` is ``None``);
+        otherwise False.
+    """
+    if bounds is None:
+        return True
+    if len(bounds) != theta.size:
+        raise ValueError(f"bounds length {len(bounds)} != theta length {theta.size}")
```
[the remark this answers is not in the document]
nikosarcevic: 243daefdf750f13484c4413d14c3786d003cf142

src/derivkit/utils/numerics.py
nikosarcevic: 2862d5cb6363675cc3d529656610acaf8dc23079

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+    if arr.ndim != 1:
+        raise ValueError(f"{name} must be 1D, got shape {arr.shape}")
+    return arr.astype(np.float64, copy=False)
+
+
+def get_index_value(theta: ArrayLike, index: int, *, name: str = "theta") -> float:
+    """Return theta[index] as a float with a clear bounds error.
+
+    This helper is useful for one-dimensional priors/likelihood pieces that
+    operate on a single parameter component.
```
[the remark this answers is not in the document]
nikosarcevic: aff6997de365ba766232ab9a7c50dd3ad94bb962

src/derivkit/forecasting/priors/core.py
nikosarcevic: a6212839af156a95b5619a122e2012eba618138c

src/derivkit/forecasting/priors/core.py
nikosarcevic: Even though it is simple, keeping prior_none makes the API explicit and self-documenting. It provides a named constructor for the “no prior / flat prior” case, which is useful for readability, consistency with other prior_* helpers, and future extensibility. It also avoids requiring users to special-case None in user code.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+        raise ValueError(f"means must be (n, p), got shape {mus.shape}")
+    n, p = mus.shape
+    if th.size != p:
+        raise ValueError(f"theta length {th.size} != p {p}")
+    if inv_covs.ndim != 3 or inv_covs.shape != (n, p, p):
+        raise ValueError(f"inv_covs must be (n, p, p) with n={n}, p={p}, got shape {inv_covs.shape}")
+    if lw.ndim != 1 or lw.size != n:
+        raise ValueError(f"log_weights must be (n,), got shape {lw.shape}")
+    if lcn.ndim != 1 or lcn.size != n:
+        raise ValueError(f"log_component_norm must be (n,), got shape {lcn.shape}")
```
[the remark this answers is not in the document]
nikosarcevic: 5d564513a61de38cd818dd3326434e402402ade6

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    "gaussian": prior_gaussian,
+    "gaussian_diag": prior_gaussian_diag,
+    "log_uniform": prior_log_uniform,
+    "jeffreys": prior_jeffreys,
+    "half_normal": prior_half_normal,
+    "half_cauchy": prior_half_cauchy,
+    "log_normal": prior_log_normal,
+    "beta": prior_beta,
+    "gaussian_mixture": prior_gaussian_mixture,
+}
```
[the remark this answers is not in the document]
nikosarcevic: I’d prefer to keep the registry here for now since it is tightly coupled to the prior constructors and only used by build_prior. Splitting it into a separate file would add indirection without much benefit at the moment. If the registry grows or needs to be shared more broadly in the future, we can revisit moving it out.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+        raise ValueError(f"log_weights must be (n,), got shape {lw.shape}")
+    if lcn.ndim != 1 or lcn.size != n:
+        raise ValueError(f"log_component_norm must be (n,), got shape {lcn.shape}")
+
+    vals = np.empty(n, dtype=float)
+    for i in range(n):
+        d = th - mus[i]
+        vals[i] = lw[i] + lcn[i] - 0.5 * float(d @ inv_covs[i] @ d)
+
+    return float(logsumexp_1d(vals))
```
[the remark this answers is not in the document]
nikosarcevic: the lhood and the prior serve different conceptual roles, even when they happen to share a gaussian functional form. 

likelihood represents information coming from the data and its noise model, while the prior encodes external knowledge, assumptions on the parameters themselves. so reusing likelihood gaussains for priors would blur this separation of responsibilities, introduce unnecessary coupling between the likelihood and prior APIs, and make the code harder to reason about and maintain. 

keeping gaussian priors in the priors module makes their purpose explicit, keeps the construction self-contained, and preserves a clear conceptual boundary between data-driven and assumption-driven components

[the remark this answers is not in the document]
nikosarcevic: also, the existing likelihood gaussian is a grid-evaluation utility (grids, pdf/logpdf), while priors here are pointwise terms for sampling/inference: logprior(theta) returns  float (or -np.inf) so reusing the likelihood implementation would force an awkward mismatch in return types and responsibilities liek i said

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    mu = np.asarray(mean, dtype=float)
+    sig = np.asarray(sigma, dtype=float)
+
+    if mu.ndim != 1 or sig.ndim != 1 or mu.shape != sig.shape:
+        raise ValueError("mean and sigma must be 1D arrays with the same shape")
+    if np.any(sig <= 0.0):
+        raise ValueError("all sigma entries must be > 0")
+
+    inv_cov = np.diag(1.0 / (sig ** 2))
+    return partial(_prior_gaussian_diag_impl, mean=mu, inv_cov=inv_cov)
```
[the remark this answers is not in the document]
nikosarcevic: answered above

src/derivkit/forecasting/priors/core.py
nikosarcevic: b620de8a759b0ce88f132e51bdd9231c8c001234

src/derivkit/forecasting/priors/core.py
nikosarcevic: removed in b620de8a759b0ce88f132e51bdd9231c8c001234
what this does is checks bascialy that  _prior_1d_impl is an internal entry point, as callers may
bypass the public constructors and partially apply it directly.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
+"""Prior utilities (core priors + unified builder).
+
+Priors are represented as callables:
+
+    logprior(theta) -> float
+
```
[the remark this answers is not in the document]
nikosarcevic: i tihnk i added more stuff here 02bf5c84987fea48854d1290ef6a73c16f2c5d61

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    inv_cov = np.diag(1.0 / (sig ** 2))
+    return partial(_prior_gaussian_diag_impl, mean=mu, inv_cov=inv_cov)
+
+
+def prior_log_uniform(*, index: int) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a log-uniform prior for a single positive parameter.
+
+    This prior has density proportional to 1/x for x > 0, which means it is
+    uniform when expressed in log(x). It is commonly used for scale parameters
+    when equal weight per order of magnitude is desired.
```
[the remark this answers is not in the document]
nikosarcevic: 02bf5c84987fea48854d1290ef6a73c16f2c5d61

src/derivkit/forecasting/priors/core.py
nikosarcevic: 370be16559228cbed9dbdd6b7b091a524de9a9f7

src/derivkit/forecasting/priors/core.py
nikosarcevic: 066c362bc68028449d7c07b1b1b45f319c512ffe

src/derivkit/forecasting/priors/core.py
nikosarcevic: see above where you asekd similar

src/derivkit/utils/numerics.py
```diff
@@ -56,3 +66,242 @@ def relative_error(a: np.ndarray, b: np.ndarray) -> float:
...
+
+    Raises:
+        ValueError: If no terms are provided.
+    """
+    if len(terms) == 0:
+        raise ValueError("sum_terms requires at least one term")
+    return partial(_sum_terms_impl, terms=terms)
+
+
+def as_1d_float_array(x: ArrayLike, *, name: str = "x") -> NDArray[np.float64]:
```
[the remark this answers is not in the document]
nikosarcevic: I responded to this and i do not see it now so ill jsut say we need it

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    "beta": prior_beta,
+    "gaussian_mixture": prior_gaussian_mixture,
+}
+
+
+def _make_prior_term(spec: dict[str, Any]) -> Callable[[NDArray[np.floating]], float]:
+    """Builds a single prior term from a configuration dictionary.
+
+    Expected format:
+        {"name": "<prior_name>", "params": {...}, "bounds": optional_bounds}
```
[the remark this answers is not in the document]
nikosarcevic: 85598c5c575684f0f05cd76d828397a9ad96bfae

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+        spec: Prior specification dictionary.
+
+    Returns:
+        Callable log-prior term: logp(theta) -> float.
+
+    Notes:
+        - For name="uniform", bounds must be provided via params={"bounds": ...}.
+          The top-level "bounds" key is reserved for optional *extra* hard bounds
+          on non-uniform terms.
+    """
```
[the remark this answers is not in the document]
nikosarcevic: all fied in this ea16018991fd08e422330d4d468dd3a76b8ae165

src/derivkit/forecasting/priors/core.py
nikosarcevic: id prefer to keep the early return here. The uniform case is a genuinely different construction (the bounds define the prior), and forcing a single return would require mutating state or adding flags to avoid double-applying bounds, which is less clear.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+    name = str(spec.get("name", "")).strip().lower()
+    if not name:
+        raise ValueError("prior spec must include non-empty 'name'")
+    if name not in _PRIOR_REGISTRY:
+        raise ValueError(f"Unknown prior name '{name}'")
+
+    params = dict(spec.get("params", {}))
+    term_bounds = spec.get("bounds", None)
+
+    if name == "uniform":
```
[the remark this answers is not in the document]
nikosarcevic: yes

for every other prior, params fully define the distribution and bound (if they present) is just an additional hard cutoff layered on top.
in case of uniform prior there is no meaningful distribuiton without bounds and allowng bounds in two places (params["bounds"] vs top-level "bounds") is purely about user-facing convenience and avoiding ambiguity—so it needs a small special-case to (1) require bounds and (2) reject “bounds provided twice

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,762 @@
...
+      build_prior(bounds=[(0.0, None), (None, None)])
+      build_prior(terms=[("gaussian_diag", {"mean": mu, "sigma": sig})])
+      build_prior(
+          terms=[
+              ("gaussian_diag", {"mean": mu, "sigma": sig}),
+              ("log_uniform", {"index": 0}),
+              {"name": "beta", "params": {"index": 2, "alpha": 2.0, "beta": 5.0}, "bounds": [(0,1), (None,None), (0,1)]},
+          ],
+          bounds=[(0.0, None), (None, None), (0.0, 1.0)],
+      )
```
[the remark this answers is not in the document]
nikosarcevic: f76a305f73b6fea3c00d36abc01a0296e622e651

src/derivkit/forecasting/priors/core.py
nikosarcevic: already explained above

src/derivkit/forecasting/priors/core.py
nikosarcevic: done here 2cb353301969233717ce98c358da35b00c4ed090 i think

src/derivkit/forecasting/priors/core.py
nikosarcevic: 2cb353301969233717ce98c358da35b00c4ed090

src/derivkit/forecasting/priors/core.py
nikosarcevic: how do you liek this version 74713a9028c52c0395c8f715920bee75b393637d

src/derivkit/utils/numerics.py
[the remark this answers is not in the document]
lonbar: That commit describes a change in a different file.

src/derivkit/utils/numerics.py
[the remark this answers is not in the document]
lonbar: It is defined in line 212 and used in 254.

src/derivkit/forecasting/priors/core.py
lonbar: This function still contains the assignment `_ = theta`. I understand that this is to circumvent warnings in pycharm, but I still think that this isn't good practice. Perhaps it would make sense to write the function parameter as `_theta` (or maybe `_` instead? That should indicate to pycharm that it is a dummy parameter. In other words, the function signature could be something like
```
def _prior_none_impl(_theta: NDArray[np.floating]) -> float:
```
and the assignment
```
_ = theta
```
can be removed.

src/derivkit/forecasting/priors/core.py
lonbar: I had a general comment that every prior function has an `_impl` companion and that it wasn't clear from the PR why these are necessary. There is no reason why the logic of the companion functions cannot be moved to the prior functions themselves, unless I am missing something obvious. In that case I'd like to understand why they are needed.

src/derivkit/forecasting/priors/core.py
lonbar: @nikosarcevic pinging you in case you missed it.

src/derivkit/forecasting/priors/core.py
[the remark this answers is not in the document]
lonbar: that commit does not seem related.

src/derivkit/forecasting/priors/core.py
lonbar: @nikosarcevic pinging you in case you missed it.

src/derivkit/forecasting/priors/core.py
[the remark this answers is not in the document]
lonbar: That is markdown code block syntax. We're using rST syntax. I'll make a separate note for this.

src/derivkit/forecasting/priors/core.py
lonbar: This commit doesn't address this particular line.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,851 @@
+"""Prior utilities (core priors + unified builder).
+
+Priors are represented as callables:
+
+    ``logprior(theta) -> float``
```
lonbar:
```suggestion
Priors are represented as callables::

    logprior(theta) -> float`
```
The formatting as-is is markdown. We're using rST.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,851 @@
...
+    means: NDArray[np.floating],
+    inv_covs: NDArray[np.floating],
+    log_weights: NDArray[np.floating],
+    log_component_norm: NDArray[np.floating],
+) -> float:
+    """Evaluates a Gaussian mixture log-prior at ``theta``.
+
+    This function computes the log of a weighted sum of Gaussian components:
+
+        ``p(theta) = sum_n w_n * N(theta | means_n, cov_n)``
```
lonbar:
```suggestion
    This function computes the log of a weighted sum of Gaussian components::

        p(theta) = sum_n w_n * N(theta | means_n, cov_n)

```

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,851 @@
...
+        *,
+        index: int,
+        sigma: float | np.floating,
+) -> Callable[[NDArray[np.floating]], float]:
+    """"Constructs a half-normal prior for a single non-negative parameter.
+
+    This prior is a standard weakly informative choice for non-negative amplitude
+    or scale parameters. It is conceptually different from a normal prior with a
+    non-negativity bound, which corresponds to a truncated normal. The half-normal
+    instead arises as the distribution of ``|N(0, sigma)|``.
```
lonbar: Please clarify what `N` means.

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,851 @@
...
+    weights: NDArray[np.floating] | None = None,
+    log_weights: NDArray[np.floating] | None = None,
+    include_component_norm: bool = True,
+) -> Callable[[NDArray[np.floating]], float]:
+    """Constructs a Gaussian mixture prior (up to an additive constant).
+
+    This prior has density proportional to a weighted sum of Gaussian components.
+
+    The mixture is:
+        ``p(theta) = sum_n w_n * N(theta | mean_n, cov_n)``
```
lonbar:
```suggestion
    The mixture is:

        p(theta) = sum_n w_n * N(theta | mean_n, cov_n)

```

src/derivkit/forecasting/priors/core.py
```diff
@@ -0,0 +1,851 @@
...
+
+    Args:
+        means: Component means with shape (n, p).
+        covs: Component covariances with shape (n, p, p).
+        inv_covs: Component inverse covariances with shape (n, p, p).
+        weights: Mixture weights with shape (n,). Can include zeros.
+        log_weights: Log-weights with shape (n,). Can include -inf entries.
+        include_component_norm: If True (default), include the per-component
+            Gaussian normalization factor proportional to |C_n|^{-1/2}.
+            This is important for *mixtures* when covariances differ.
```
lonbar:
```suggestion
        means: Component means with shape ``(n, p)``.
        covs: Component covariances with shape ``(n, p, p)``.
        inv_covs: Component inverse covariances with shape ``(n, p, p)``.
        weights: Mixture weights with shape ``(n,)``. Can include zeros.
        log_weights: Log-weights with shape ``(n,)``. Can include -inf entries.
        include_component_norm: If ``True`` (default), include the per-component
            Gaussian normalization factor proportional to :math:`|C_n|^{-1/2}`.
            This is important for *mixtures* when covariances differ.
```

lonbar (changes requested):
Found some rST formatting issues that will prevent the docs from being built.

src/derivkit/forecasting/priors/core.py
nikosarcevic: by “constructor” I mean the public prior_* functions that return a logprior(theta) callable.


partially apply it directly refers to using functools.partial(_prior_1d_impl, ...) as a prior instead of going through a prior_* helper. That’s possible since _prior_1d_impl is a module-level function
the extra validation was meant to make that safe -- it’s now removed since it wasn’t necessary and the comment was confusing.

src/derivkit/forecasting/priors/core.py
nikosarcevic: chanaged this everywhere in the script f40c0094ecec344ed1c7488c7f1ceecc7ca4859e

src/derivkit/forecasting/priors/core.py
nikosarcevic: soz again it was formatted properly https://github.com/derivkit/derivkit/blob/fa8fa8a514e18a99dc34e3793935fce09334c93b/src/derivkit/forecasting/priors/core.py#L346

src/derivkit/forecasting/priors/core.py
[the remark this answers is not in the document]
lonbar: I'm still not sure I agree with this, but in the interest of moving things along we can go with your preference.

src/derivkit/forecasting/priors/core.py
lonbar: @ctrendafilova just pinging you on this.

Issue #306, opened 2025-12-16
End-to-end forecasting: from Fisher/DALI tensors to posteriors and contours


nikosarcevic:
DerivKit currently outputs forecast tensors (Fisher, DALI), not posteriors.

To make contours, users must define conventions (delta chi2 vs log-posterior), apply priors, and compute intermediate products themselves.
This extension standardizes those missing steps so tensors become samplable posteriors, with explicit conventions and optional GetDist integration.

Without this, users must reimplement core logic and the workflow feels incomplete.
The work is split into small PRs (priors, scalar objectives, sampling, GetDist) to keep reviews manageable.

This needs PRs in steps
 - priors
 - intermediate products (delta chi2, logposterior code)
 - and finally a getdist integration

nikosarcevic:
this can be closed as we merged #305 #307 #308 and #309

PR #309, opened 2025-12-16
GetDist Integration PR4 - Add modified forecasting tests

nikosarcevic:
This PR builds on top of PR3 and slightly modifies some of the forecasting tests
Also adds a necessary toml changes to include getdsit and emcee 

This is the last step in closing #306

### DIFF
--- a/tests/test_forecast_integrations_getdist_fisher.py
+++ b/tests/test_forecast_integrations_getdist_fisher.py
@@ -24,8 +24,6 @@ def test_fisher_to_getdist_gaussiannd_mean_and_cov_and_names_and_label():
     assert list(g.names) == names
     assert g.label == "X"
 
-    # Labels: GaussianND doesn't expose `.labels` in this getdist version.
-    # Check via the parameter names/labels container if present.
     if hasattr(g, "paramNames") and g.paramNames is not None:
         assert [p.name for p in g.paramNames.names] == names
         assert [p.label for p in g.paramNames.names] == labels


tests/test_forecast_integration_getdist_dali.py
```diff
@@ -0,0 +1,237 @@
...
+import pytest
+
+from derivkit.forecasting.integrations.getdist_dali_samples import (
+    dali_to_getdist_emcee,
+    dali_to_getdist_importance,
+)
+
+
+def _toy_dali_inputs(p: int = 2):
+    """Small consistent (theta0, fisher, G, H) for testing."""
```
lonbar: Please elaborate a bit. This creates dummy DALI data?

tests/test_forecast_integrations_getdist_fisher.py
```diff
@@ -0,0 +1,207 @@
...
+    labels = [r"p_0", r"p_1"]
+
+    g = fisher_to_getdist_gaussiannd(theta0, fisher, names=names, labels=labels, label="X")
+
+    assert np.allclose(np.asarray(g.means[0]), theta0)
+    assert np.allclose(np.asarray(g.covs[0]), fisher_to_cov(fisher))
+    assert list(g.names) == names
+    assert g.label == "X"
+
+    # Labels: GaussianND doesn't expose `.labels` in this getdist version.
```
lonbar: Please clarify "this version". We don't currently specify a version constraint for getdist. If this is necessary we will have to update pyproject.toml.

nikosarcevic:
> Looks comprehensive. Added two minor nits.
> 
> One thing that may be missing is that it doesn't always seem immediately clear why the expected output is expected. We could add a few more comments clarifying this. I don't think that is unique to the added tests; probably we could improve the test suite overall with this. So it may not be necessary to do this now, but it would be good to think about it. What do you think?

yeha it is of course a good idea but I am so done with making millions of tests so we can perhaps postpone it for later? or if you or @ctrendafilova wanna pick it up.

tests/test_forecast_integration_getdist_dali.py
nikosarcevic: yes. i have clarified in 1343807fe0a4b61cc4cf67768b1252cb341e0739

I also had to change imports in these test files since i have merged main into this. all tests pass

lonbar (approved):
MRDZ

PR #313, opened 2025-12-25
Add opt-in JAX autodiff helpers + tests

nikosarcevic:
This PR adds an experimental, opt-in JAX autodiff module to DerivKit.

Includes helpers for derivatives, gradients, Jacobians, and Hessians with clear failure modes when JAX is unavailable or functions are not differentiable.
No backend is registered by default; users must explicitly enable it.

This is useful to have for our benchmarking (proving a point that no deriv methid is perfect)

### DIFF
--- a/docs/derivkit.autodiff.jax_autodiff.rst
+++ b/docs/derivkit.autodiff.jax_autodiff.rst
@@ -1,0 +1,7 @@
+derivkit.autodiff.jax\_autodiff module
+======================================
+
+.. automodule:: derivkit.autodiff.jax_autodiff
+   :members:
+   :show-inheritance:
+   :undoc-members:
--- a/docs/derivkit.autodiff.jax_core.rst
+++ b/docs/derivkit.autodiff.jax_core.rst
@@ -1,0 +1,7 @@
+derivkit.autodiff.jax\_core module
+==================================
+
+.. automodule:: derivkit.autodiff.jax_core
+   :members:
+   :show-inheritance:
+   :undoc-members:
--- a/docs/derivkit.autodiff.jax_utils.rst
+++ b/docs/derivkit.autodiff.jax_utils.rst
@@ -1,0 +1,7 @@
+derivkit.autodiff.jax\_utils module
+===================================
+
+.. automodule:: derivkit.autodiff.jax_utils
+   :members:
+   :show-inheritance:
+   :undoc-members:
--- a/docs/derivkit.autodiff.rst
+++ b/docs/derivkit.autodiff.rst
@@ -1,0 +1,20 @@
+derivkit.autodiff package
+=========================
+
+Submodules
+----------
+
+.. toctree::
+   :maxdepth: 4
+
+   derivkit.autodiff.jax_autodiff
+   derivkit.autodiff.jax_core
+   derivkit.autodiff.jax_utils
+
+Module contents
+---------------
+
+.. automodule:: derivkit.autodiff
+   :members:
+   :show-inheritance:
+   :undoc-members:
--- a/docs/derivkit.rst
+++ b/docs/derivkit.rst
@@ -8,6 +8,7 @@ Subpackages
    :maxdepth: 4
 
    derivkit.adaptive
+   derivkit.autodiff
    derivkit.calculus
    derivkit.forecasting
    derivkit.likelihoods


nikosarcevic:
> Quick initial question: If it is meant as an experiment it doesn't need to be added to the main branch, but can live in its own branch, right? It shouldn't be difficult to keep it in sync with the main branch, as it adds a new subpackage.

it should live on main but NOT in official DerivativeKit.
I do not think it is reasonable to ask people to "just go to this branch --- then you clone then you locally install..." it is too much overhead when it can simple exist on main. And I even added an opt-in option to register it w DerivativeKit when you run so it is just easy.

src/derivkit/autodiff/jax_utils.py
```diff
@@ -0,0 +1,147 @@
+"""Utilities for JAX-based autodiff in DerivKit."""
+
+from __future__ import annotations
+
+from typing import Any, Callable
+
+try:
+    import jax
+    import jax.numpy as jnp
+    from jax.errors import ConcretizationTypeError
```
lonbar: `jax` must be added as a dependency. Since you don't want it to be available by default I'd suggset that we add this as an optional dependency in pyproject.toml.

tests/test_autodiff_jax_autodiff.py
```diff
@@ -0,0 +1,76 @@
+"""Unit tests for JAX-based automatic differentiation engine integration."""
+
+from __future__ import annotations
+
+import pytest
+
+pytest.importorskip("jax")  # noqa: E402
```
lonbar: This line should not be needed as jax is a depedency

tests/test_autodiff_jax_core.py
```diff
@@ -0,0 +1,129 @@
+"""Unit tests for derivkit.autodiff.jax_core functions."""
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+pytest.importorskip("jax")
+import jax.numpy as jnp  # noqa: E402
```
lonbar: This should not be needed as jax is a dependency

src/derivkit/autodiff/jax_autodiff.py
```diff
@@ -0,0 +1,99 @@
...
+    ...     return jnp.sin(x) + 0.5 * x**2
+    ...
+    >>> dk = DerivativeKit(func, 1.0)
+    >>> dk.differentiate(method="autodiff", order=1)
+    1.0...
+    >>> dk.differentiate(method="autodiff", order=2)
+    0.5...
+
+Notes:
+-----
```
lonbar: Either the colon must be removed or you need to add a final `-`.

src/derivkit/autodiff/jax_autodiff.py
```diff
@@ -0,0 +1,99 @@
...
+- This backend is scalar-only. For gradients/Jacobians/Hessians of functions
+  with vector inputs/outputs, use the standalone helpers in
+  ``derivkit.autodiff.jax_core`` (e.g. ``autodiff_gradient``).
+- JAX is an optional dependency. If ``jax``/``jaxlib`` are not installed,
+  ``register_jax_autodiff_backend`` will raise an error.
```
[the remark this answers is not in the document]
lonbar: > - JAX is an optional dependency. If ``jax``/``jaxlib`` are not installed,
  ``register_jax_autodiff_backend`` will raise an error.

Since it is an optional dependency it should be installed by the user, so this is redundant.

src/derivkit/autodiff/jax_core.py
```diff
@@ -0,0 +1,209 @@
...
+for an opt-in integration.
+
+Use only with JAX-differentiable functions. For arbitrary models, prefer
+the "adaptive" or "finite" methods.
+
+Shape conventions (aligned with DerivKit calculus builders):
+- autodiff_derivative: f: R -> R                      -> scalar float
+- autodiff_gradient:   f: R^n -> R                    -> (n,)
+- autodiff_jacobian:   f: R^n -> R^m (or tensor)      -> (m, n) with m = prod(out_shape)
+- autodiff_hessian:    f: R^n -> R                    -> (n, n)
```
lonbar: This will need better formatting.

lonbar:
@nikosarcevic I have added a quick review with mostly minor comments. The code seems a bit awkward to me, in a large part because it can't make up its mind whether or not jax is available or not. If this is added to derivkit then jax must be listed as a dependency in pyproject.toml. Since you seem to not want this to be available by default I'd suggest to add it as an optional dependency, so the user can make it available via e.g.
```
pip install derivkit[autodiff]
```
In that case jax can be expected to be available for the user, though this must be documented somewhere.

Havind said that—
> I do not think it is reasonable to ask people to "just go to this branch --- then you clone then you locally install..." it is too much overhead when it can simple exist on main. And I even added an opt-in option to register it w DerivativeKit when you run so it is just easy.

I think that is fair, but as I mentioned above it should still be documented somewhere. It is okay to add something and have a note in the docs saying "look, this is experimental, so no guarantees". Jax should still be available as an optional dependency.

src/derivkit/autodiff/jax_utils.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Raises:
+        AutodiffUnavailable: If JAX is not installed.
+    """
+    if not _HAS_JAX:
+        raise AutodiffUnavailable(
+            "JAX autodiff requires `jax` + `jaxlib`.\n"
+            "Install them with `pip install jax jaxlib`."
+        )
+
```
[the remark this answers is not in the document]
nikosarcevic: done in cfdead14ff7f550515a2c3475981591e9b501e73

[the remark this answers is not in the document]
nikosarcevic: i have also added several other commits related to this

- isntall guidance 320da1acc8e3a662cad3f0824087b2493b64fcaa
- failure handling 8b188a0b3e3101aec69fa9f5f4b89cc41bdbd312
- docs 3a462cd2c8d2decb81f80c9b6d9a30e35ea9950b in install
- docs 17ef1f70d45bad1004f09fd72cc1d3ddd9f908dc in guide

src/derivkit/autodiff/jax_utils.py
```diff
@@ -0,0 +1,147 @@
...
+def to_jax_scalar(y: Any, *, where: str) -> "jnp.ndarray":
+    """Ensures that output is scalar and returns as JAX array.
+
+    Args:
+        y: Output to check.
+        where: Context string for error messages.
+
+    Returns:
+        JAX array with shape ().
```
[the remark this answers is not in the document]
nikosarcevic: good point — clarified the docstring to explicitly describe this as a scalar (0-d) jax array  and made the corresponding  wording explicit for array outputs as well in 196ef5642bf5f4e4159dabbb1b7fee3cb15ea521

tests/test_autodiff_jax_autodiff.py
nikosarcevic: Removed the redundant pytest.importorskip("jax") and rely on require_jax() + pytest.skip instead

since jax is an optional extra (not installed in the default test env), the tests still need to skip cleanly when it’s not present ef696fced14844eb4b374c164e4076374cf9a412

tests/test_autodiff_jax_core.py
nikosarcevic: Agreed — removed pytest.importorskip("jax")

to keep ruff happy (no local imports) while still treating jax as optional, I basically gate the jax.numpy import behind require_jax() at module import time and skip tests via require_jax()/pytest.skip when unavailable 
097fc4442715a6bc4a79a3ab6cdb07c564f593d4

tests/test_autodiff_jax_utils.py
```diff
@@ -0,0 +1,65 @@
+"""Unit tests for derivkit.autodiff.jax_utils module."""
+
+from __future__ import annotations
+
+import pytest
+
+pytest.importorskip("jax")
+import jax.numpy as jnp  # noqa: E402
```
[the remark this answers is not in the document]
nikosarcevic: already explained what we need to do (if we are keeping jax opt-in) but i did some minor changes here i tihnk oyu will approve 06a2f7c72e0961bf1b49fa6e7ea5661acb2c9298

src/derivkit/autodiff/jax_autodiff.py
nikosarcevic: 8e6e3448b49ae6d8dc55c3af9db5eb2d1e7662fb

PR #323, opened 2025-12-30
Docs/examples structure

### LINKED ISSUE #321: Docs: Organize and add usage examples
We already have in-depth DerivKit demo scripts, which are awesome for full workflows, but the docs themselves are currently very API-heavy.

This issue proposes adding a small Examples section in the Docs with minimal, task-oriented usage snippets that sit between the API reference and the full demos.

Goal: quickly answer “how do I do X?” without sending users straight to long scripts.

Strucutre:
Derivatives (different methods: finite diff, adaptive/local poly, fornberg, tabulated)
Calculus (gradient / jacobian / hessian)
Forecasting (Fisher, Fisher bias, DALI)

no API changes, not replacing demos.

Implementation: add docs/examples/index.rst + category pages + link Examples from docs/index.rst.

nikosarcevic:
This PR adds super simple exampels to docs

Closes #321

### DIFF
--- a/docs/examples/forecasting/dali_getdist.rst
+++ b/docs/examples/forecasting/dali_getdist.rst
@@ -1,5 +1,5 @@
 DALI sampling and GetDist integration
-====================================
+=====================================
 
 DerivKit can convert a DALI-expanded posterior into GetDist-ready objects for
 plotting (triangle/corner plots) and lightweight downstream analysis.


nikosarcevic:
> This looks great! Thanks for adding this.
> 
> I haven't tried to run any of the code snippets; I trust that you took care that these are valid. One thing to keep in mind is that now we're starting to approach the point where we may want to be careful that example code snippets are valid. I think there is a a package for this, but I can't recall it right now, and don't have the time to look into it right now. Maybe you want to do this for this PR, or do you think it makes more sense to do this later?
> 
> I have two minor comments which apply to this PR in general:
> 
> 1. In at least one point you tell the reader to "check the API docs". It would be helpful to provide a link to the main page of the API reference.
> 2. The examples are provided as code blocks, which I suppose are meant for the user to copy and paste. The user can then check the output, which is not provided in the code. Do we want to instead provide an interpreter like display which shows both the user input and standard output? I'm guessing but not sure that if we want to add the "examples checker" I mentioned above that we want this output to be present.

Thanks.

Yeah those are great suggestions. Let us decide this tonight. I will fix the refs and then definitely we should see how to deal with your suggestion 2 (I am for it, thank you!)

FYI not all will work currently until we have merged the getdist integration. After that, it should work (I was checkign locally with the local installaton of what is currently PR3)

PR #336, opened 2026-01-02
PR2 (tests): Tests for generalized Fisher + X–Y Fisher

nikosarcevic:
Adds unit tests covering the new generalized Gaussian Fisher and X–Y Fisher functionality, plus updates to existing forecasting tests to exercise the new API paths.

Builds on PR1


nikosarcevic:
i had to change indeixng -- 
you were indexing the covariance derivatives along the wrong axis, effectively mixing up which matrix corresponded to which parameter, so the code was not using the same dC/dtheta_i objects as in the paper

i reverted it back to what it was -- computes it by reshaping the derivatives so that each parameter’s covariance derivative is handled separately and correctly

Issue #352, opened 2026-01-04
Make Poisson likelihood return log-likelihood by default


nikosarcevic:
LikelihoodKit.poissonian currnetly defaults to returning PMF values, while the Gaussian likelihood defaults to log-likelihoods. This is inconsistent and can lead to numerical instabilities  for some moerdate cout values.

so we need to change the default of poissonian to return_log=True, keeping PMF output available via return_log=False.

we NEED to do this since log-likelihoods are the natural object for inference and likelihood combination, and this matches both the Gaussian API and the documentation.

NOTE -- change the likelihoods exampels accordingly. those are currently in review in PR C - Feature/docs likelihood examples 347 #349

PR #353, opened 2026-01-04
Update default value of `return_log` in likelihood_kit and likelihood functions

lonbar:
Solves #352.

### DIFF
--- a/src/derivkit/likelihoods/poisson.py
+++ b/src/derivkit/likelihoods/poisson.py
@@ -36,7 +36,8 @@ def build_poissonian_likelihood(
     Args:
         data: an array representing the given data values.
         model_parameters: an array representing the means of the data samples.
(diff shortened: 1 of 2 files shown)


src/derivkit/likelihoods/poisson.py
```diff
@@ -36,8 +36,7 @@ def build_poissonian_likelihood(
     Args:
         data: an array representing the given data values.
         model_parameters: an array representing the means of the data samples.
-        return_log: when set to ``True``, returns the log-likelihood. Defaults
-            to ``False``.
+        return_log: when set to ``True``, returns the log-likelihood.
```
nikosarcevic: write what is otherwise

nikosarcevic: also see in other places in docstring where this needs to be explained liek if true this if false this

PR #359, opened 2026-01-06
Feature/tox cross platform docs

### LINKED ISSUE #358: Make tox -e do cross-platform and tidy pyproject.toml and conf.py
tox -e do currently works only on macOS because it uses the open command to launch the built HTML docs. This fails on Linux (and Windows).

conf py is alos a bit messy now and that should be clean and readable.

We should:

- make the “build + open docs” workflow cross-platform (e.g. use Python webbrowser instead of open)
- keep tox -e docs fully portable and CI-safe
- clean up the docs-related tox envs in pyproject.toml (reduce duplication, clearer naming)
- clean up conf so it is readable

Goal: docs tox envs work on macOS and Linux, and the TOML stays readable.

nikosarcevic:
In this PR I have
- cleaned up pyproject.toml
- cleaned up conf.py in docs
- added cross platform support for the convenient shorthand `tox -e do` (builds docs, runs doctest and opens the html locally)

this closes #358

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -181,6 +181,7 @@ commands = [
 [tool.tox.env.docs-releases]
 description = "Build the documentation for derivkit releases using Sphinx"
 extras = ["docs"]
+allowlist_externals = ["cp"]
 commands = [
   [
     "sphinx-apidoc",


docs/conf.py
```diff
@@ -1,29 +1,35 @@
...
 
-# For the full list of built-in configuration values, see the documentation:
-# https://www.sphinx-doc.org/en/master/usage/configuration.html
-
-# -- Project information -----------------------------------------------------
-# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
-
+# -----------------------------------------------------------------------------
+# Project information
+# -----------------------------------------------------------------------------
```
lonbar: I would prefer to keep this as it was, as it gives the links to where we can find more information.

docs/conf.py
```diff
@@ -38,34 +44,34 @@
     "sphinx_copybutton",
```
lonbar:
```suggestion
```
This plugin can't be used with github pages.

pyproject.toml
```diff
@@ -17,74 +39,57 @@ dependencies = [
...
-"test" = [
-    "pytest",
+test = ["pytest"]
+
+docs = [
+  "sphinx",
+  "sphinx-design",
+  "sphinx-multiversion",
+  "furo",
+  "sphinx-copybutton",
```
lonbar:
```suggestion
```

pyproject.toml
```diff
@@ -17,74 +39,57 @@ dependencies = [
...
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
+
 [tool.tox]
 requires = ["tox>=4.22"]
 labels = { test = ["py310", "py311", "py312", "py313"] }
 envlist = [
   "docs",
+  "do",
```
lonbar:
```suggestion
  "do",
  "docs",
```

pyproject.toml
```diff
@@ -93,132 +98,68 @@ envlist = [
...
-  [
-    "sphinx-build",
-    "-b", "html",
-    "{tox_root}/docs",
-    "{tox_root}/docs/_build/html",
-    "--fail-on-warning",
-  ],
+  ["sphinx-apidoc", "-o", "{tox_root}/docs", "{tox_root}/src/derivkit", "--separate", "-f"],
+  ["sphinx-build", "-b", "doctest", "{tox_root}/docs", "{tox_root}/docs/_build/doctest", "--fail-on-warning"],
+  ["sphinx-build", "-b", "html", "{tox_root}/docs", "{tox_root}/docs/_build/html", "--fail-on-warning"],
```
lonbar: It may be more convenient to allow a CLI flag which can be either `html` or `doctest`. In that case we won't have to run the same command twice. I don't think we need to build the documentation each time we want to run doctest, right?

pyproject.toml
```diff
@@ -93,132 +98,68 @@ envlist = [
...
 ]
 
 
-
-
 [tool.tox.env.do]
-# Same policy as `tox -e docs`, but also opens the built HTML.
-# API doctests are currently disabled via docs/conf.py exclude_patterns.
+# Local convenience: same as docs, plus open HTML index (cross-platform).
 description = "Build docs (doctest + html) and open local HTML index"
```
lonbar:
```suggestion
description = "(Mac only) Build and test docs and open docs in browser"
```

docs/conf.py
```diff
@@ -74,63 +80,72 @@
...
 #
 # To re-enable API docstring doctests later:
 #   - comment out the two lines below.
+#
 # exclude_patterns += [
-#   "modules.rst",
-#   "derivkit.*.rst",
+#     "modules.rst",
+#     "derivkit.*.rst",
 # ]
```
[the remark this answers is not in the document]
nikosarcevic: removed in f4178fa6a47906a593b32ccf1c9e4cff265485a6

pyproject.toml
```diff
@@ -93,132 +98,68 @@ envlist = [
...
-    "{tox_root}/docs",
-    "{tox_root}/docs/_build/html",
-    "--fail-on-warning",
-  ],
-  [
-    "open",
-    "{tox_root}/docs/_build/html/index.html",
+    "python",
+    "-c",
+    "import pathlib, webbrowser; p=(pathlib.Path(r'{tox_root}')/'docs'/'_build'/'html'/'index.html').resolve(); webbrowser.open(p.as_uri())",
```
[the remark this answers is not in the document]
nikosarcevic: well if im the only one using it then we can leave it

pyproject.toml
```diff
@@ -17,73 +39,60 @@ dependencies = [
...
-[tool.ruff.lint]
 # Here we list the styling/formatting rules that Ruff should use.
 # A list of available rules and what each rule does is provided at
 # https://docs.astral.sh/ruff/rules
+
+[tool.ruff]
+line-length = 79
+
+
+[tool.ruff.lint]
```
lonbar:
```suggestion
[tool.ruff]
line-length = 79


[tool.ruff.lint]
# Here we list the styling/formatting rules that Ruff should use.
# A list of available rules and what each rule does is provided at
# https://docs.astral.sh/ruff/rules
```

lonbar (approved):
I've made some final changes to pyproject.toml in aa187d9cfc87a20eb7cbbdffd0729acf4ffa9283, a34c7598aa10dca17609c64d92681910f0d0aeab, e6e08151b4d3e64d70b5ec100f950f4835b18ede, just to ensure that we wouldn't run into unexpected permissions errors once this is merged.

I think this is all good. Thanks for adding this.

PR #365, opened 2026-01-06
feature/docs missing examples 361

nikosarcevic:
Does the following:

- adds a hyperhessian example
- adds an example for Fisher-Gauss
- adds example for Fisher-xy
- updates calculus and forecasting landing pages.

### DIFF
--- a/docs/examples/forecasting/fisher_xy.rst
+++ b/docs/examples/forecasting/fisher_xy.rst
@@ -75,8 +75,6 @@ This example compares:
    ...     cxy=cxy,
    ...     cyy=cyy,
    ... )
-   >>> bool(np.allclose(fisher_std, fisher_xy, atol=1e-10, rtol=0.0))
-   False
    >>> # Convert to GetDist GaussianND samples for visualization
    >>> gnd_std = fisher_to_getdist_gaussiannd(
    ...     theta0=theta0,


docs/examples/calculus/hyperhessian.rst
```diff
@@ -0,0 +1,142 @@
...
+   >>> # Check symmetry for a representative index triplet.
+   >>> bool(hess3[0, 1, 0] == hess3[1, 0, 0] == hess3[0, 0, 1])
+   True
+
+
+Tensor-valued outputs
+---------------------
+
+If the function returns a tensor, the hyper-Hessian is computed independently
+for each output component and reshaped back to ``(*out_shape, p, p, p)``.
```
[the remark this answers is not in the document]
lonbar: However, this is a bit confusing. We've already mentioned what the output shape is, so we don't need to mention that again. Reshaping is in principle unrelated to the components being computed independently, right?

The most important thing here is the example, as it shows concretely what we mean when we say `(*out_shape, p, p, p)`.

docs/examples/calculus/hyperhessian.rst
```diff
@@ -0,0 +1,142 @@
...
+   >>> ref0[0, 0, 0] = 6.0
+   >>> np.allclose(hess3[0], ref0, atol=1e-6, rtol=0.0)
+   True
+   >>> # Component 1: all third derivatives are zero (quadratic/linear terms only).
+   >>> np.allclose(hess3[1], 0.0, atol=1e-6, rtol=0.0)
+   True
+
+
+Finite differences (Ridders)
+----------------------------
```
lonbar:
```suggestion
Tuning the derivative method
----------------------------
```
The specific method isn't that important here; we just want to show that the algorithm can be tuned with the method parameter.

docs/examples/forecasting/fisher_xy.rst
```diff
@@ -0,0 +1,223 @@
...
+   >>> # X–Y Fisher: propagate input uncertainty into an effective covariance R(theta)
+   >>> fisher_xy = build_xy_gaussian_fisher_matrix(
+   ...     theta0=theta0,
+   ...     x0=x0,
+   ...     mu_xy=mu_xy,
+   ...     cxx=cxx,
+   ...     cxy=cxy,
+   ...     cyy=cyy,
+   ... )
+   >>> bool(np.allclose(fisher_std, fisher_xy, atol=1e-10, rtol=0.0))
```
lonbar: The casting isn't wrong, but not necessary either, right?

docs/examples/calculus/hyperhessian.rst
nikosarcevic: i would leave it as is since the user should get familiar with the name of the extrapolation method. also ridders is strictly for finite

lonbar (approved):

PR #377, opened 2026-01-15
Feature/forecastkit api 338

### LINKED ISSUE #338: Expose laplace, expansions  and getdist in ForecastKit
Currently the expansions, getdist integrations, laplace approx are one level deeper in the code.

We should have the API top level wrap in forecastKit for laplace and thne for getdist integrations as well as some useful funcs from expansions.

We need to figure out how to do it (complications may arise due to the ForecastKit initialization -- perhaps some of these structures do not require all init args)

nikosarcevic:
in this PR i have
- exposed low level functions into ForecastKit so they are forecastkit methods:
  - laplace
  - expansions 
  - fisher and dali sampling
- expanded delegation test for those functions inside test_forecast_kit.py
- deleted some doc filesand added _generate citation file to gitignore since these files were left locally and are actually connected to PR #376 

this closes #338

### DIFF
--- a/src/derivkit/forecast_kit.py
+++ b/src/derivkit/forecast_kit.py
@@ -385,7 +385,7 @@ class ForecastKit:
         and quartic DALI tensors. The ``convention`` parameter controls the
         numerical prefactors applied to the cubic/quartic contractions.
 
-        The expansion point is taken from the stored ``self.theta0``.
+        The expansion point is taken from :attr:`ForecastKit.theta0`.
 
         Args:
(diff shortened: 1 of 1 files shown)


src/derivkit/forecast_kit.py
lonbar: A general comment: the distinction between fisher and DALI is no longer necessary, rigth? The fisher aproximation is the DALI singlet. So while we may want to keep separate wrappers for people's convenience, those wrappers can call the same underlying function

PR #383, opened 2026-01-27
Store intermediate orders for full forecast

lonbar:
This is an improvement of `get_forecast_tensors`. Since, for a forecast of DALI order `n`, all derivatives up to order `n` have to be computed anyway there is no reason not to determine and return all forecast tensors up to the given `forecast_order`.

That is, if the user requests _e.g._ a forecast up to DALI triplet (`forecast_order = 3`), the costs of computing the DALI singlet (the Fisher matrix) and the DALI doublet as part of the same function call are minimal. In this case the user does not have to repeatedly call `get_forecast_tensors` to compute each order separately, which can save a lot of time if the function that has to be differentiated is costly to compute.

This new feature can be activated by supplying the input `single_forecast_order=True` to `get_forecast_tensors`. The old behaviour of the function has been preserved by setting the parameter to `True`, which is its default setting.

### DIFF
--- a/tests/test_forecast_forecast_core.py
+++ b/tests/test_forecast_forecast_core.py
@@ -9,6 +9,7 @@ from derivkit.forecasting.fisher import build_delta_nu, build_fisher_bias
 from derivkit.forecasting.forecast_core import (
     _get_derivatives,
     get_forecast_tensors,
+    SUPPORTED_FORECAST_ORDERS,
 )
 from derivkit.utils.linalg import invert_covariance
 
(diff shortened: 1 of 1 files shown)


src/derivkit/forecasting/forecast_core.py
```diff
@@ -45,8 +45,13 @@ def get_forecast_tensors(
         forecast_order: int = 1,
         method: str | None = None,
         n_workers: int = 1,
+        single_forecast_order: bool = True,
```
nikosarcevic: hmm how about renaming it to include_lower_order and make it a bool?

PR #386, opened 2026-01-28
Fix/docs adoption workflows 385

nikosarcevic:
ok this hopefully solves the multiverse probelm and also hopefully solves the workflow issue.

i cnanot test it like raise issue -- automoatically get a pr, review -- shows up in the docs. we need to test this.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -154,8 +154,12 @@ html_sidebar = {
 # -----------------------------------------------------------------------------
 # Sphinx Multiversion
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/derivkit-adoption.yml
lonbar: why are these changes necessary? Isn't it useful to enforce a validation?

docs/conf.py
lonbar: These changes do not address the problem of the CI, so we should leave them out.

PR #397, opened 2026-01-30
Fix/docs c suggestions 394

### LINKED ISSUE #394: fix docs C suggestions
Cynthia went thru the docs and has some suggestions. These suggestions are in the google docs currently so i am copying them here verbatim

About ✅
Installation ✅
Examples ✅
Workflows ✅
derivkit ✅
Citation ✅
Contributing ✅
Team ✅
License ✅

https://docs.derivkit.org/main/about/kits/derivative_kit.html

“Low overhead and easy to reason about” – What does “easy to reason about” mean?
“The method can enlarge the grid if there are too few samples, adjust the effective polynomial degree, and reports detailed diagnostics [2].” – Grammar nitpicking; I would write something more like “The method can enlarge the grid if there are too few samples and adjust the effective polynomial degree, and it reports detailed diagnostics [2].”
“JAX provides automatic differentiation for Python functions written using jax.numpy <https://docs.jax.dev/en/latest/jax.numpy.html>.” – I think the jax page is supposed to be hyperlinked?
"The function is noisy, tabulated, interpolated, or provided by an external numerical code (e.g. Boltzmann solvers or emulators in cosmology)" – All good except I’m confused by the reference to emulators. Isn’t the point of emulators to provide autodiff compatible estimates of cosmological observables?

https://docs.derivkit.org/main/about/kits/forecast_kit.html

DALI – On the derivatives page, you have some comments like "when to use this" and "when not to use this." Should we have some similar suggestions for DALI? This is on the “workflows” page. ✅
Some entries on this page are duplicated (e.g. X-Y Fisher Formalism), but the duplicates are not identical.

https://docs.derivkit.org/main/workflows.html

In the DALI section, should we add common warning points for DALI approximation, e.g. it performs poorly for sublinear models? (e.g. section VI B in 2211.06534)
“I expect non-Gaussian posteriors (banana-shaped, skewed, etc.)” – We could add a bullet point like "where physical bounds on parameters would result in non-Gaussian posteriors" or something like that. Sometimes it’s not necessarily that the mathematical model function is non-linear, but that your parameter space gets chopped off from being Gaussian. Non-infinite priors are something that is not captured with an analytic Fisher approximation (if you don't numerically sample the posterior with something like MCMC).
"I have a parameter dependent covariance, can I still use Fisher / DALI?" – "Minimal example" does not link to anything.

https://docs.derivkit.org/main/contributing.html

Typo – "environmnent" should be "environment."

https://docs.derivkit.org/main/examples/derivatives/tabulated.html

“This is useful when working with simulation outputs, precomputed theory tables, or externally generated data products where the underlying function cannot be evaluated analytically.” – I might extend the last point in the list to something like “ …  or externally generated data products where the underlying function cannot be evaluated analytically or is computationally costly to call repeatedly.”

https://docs.derivkit.org/main/examples/forecasting/fisher.html

“The diagonal elements [of the Fisher matrix] correspond to inverse variances (up to correlations).” – Wouldn’t it be more precise to say “the diagonal elements of the inverse Fisher matrix correspond to variances (up to correlations)”? If the diagonal element of a matrix has value A, it is not true that the diagonal of the inverse has value 1/A.

https://docs.derivkit.org/main/examples/forecasting/fisher_gauss.html

“The covariance dependence is chosen to be noticeable but not dominant, while remaining SPD by construction.” – What does “SPD” mean?

nikosarcevic:
This addresses all the suggestions from @ctrendafilova (many thanks!!!)

This closes #394

### DIFF
--- a/docs/about/kits/derivative_kit.rst
+++ b/docs/about/kits/derivative_kit.rst
@@ -256,13 +256,19 @@ estimating derivatives numerically, autodiff propagates
 derivatives analytically through the computational graph, yielding exact
 derivatives up to machine precision.
 
+A key distinction from finite differences or local fits is that autodiff stays
+close to the implemented functional form: it differentiates the same computation
+used to produce the model outputs, rather than estimating derivatives from
+nearby perturbed evaluations. When the model is smooth and expressed end-to-end
+in JAX, the resulting derivative typically has very small numerical error and
+does not introduce additional differencing noise.
+
 This avoids step-size choices, local sampling, and numerical differencing.
-However, this does not imply that autodiff is universally superior. A common
-misconception is that autodiff always yields a “perfect” derivative: in practice,
-it differentiates the implemented computation, which may already include
-approximations, interpolation, or numerical artifacts. As a result, autodiff can
-produce an exact derivative of an effective function that differs from the intended
-underlying model.
+This does not mean autodiff is universally superior. Autodiff differentiates the
+*implemented* computation. If that computation already contains approximations
+(splines, table lookups, iterative solvers, discontinuities, clipping, etc.),
+autodiff returns the exact derivative of that effective function, which may not
+match the derivative of the intended underlying model.
 
 Its applicability is therefore limited to fully JAX-compatible, analytic models
 and does not extend to noisy, interpolated, or externally evaluated functions.


lonbar (approved):
Looks good to me. Added some minor suggestions.
I like the superscript for the DALI multiplets.

PR #398, opened 2026-01-30
Move api reference to api subdirectory

### LINKED ISSUE #395: Docs: generate Sphinx API docs under docs/api/ instead of top-level docs/
Auto-generated Sphinx API docs are currently written into the top-level docs/ directory.
I suggest we move all generated API output into docs/api/ to keep generated content separate from hand-written documentation.

lonbar:
This includes all `derivkit*.rst` files. This should keep the root directory of the documentation more managable.

Closes #395.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -1,5 +1,5 @@
-DerivKit API reference
-======================
+derivkit API
+============
 
 .. toctree::
    :maxdepth: 2


docs/index.rst
```diff
@@ -32,7 +32,7 @@ automatic differentiation may be unavailable or inappropriate.
    installation
    examples/index
    workflows
-   modules
+   api/modules
```
nikosarcevic: can you 
- rename it so it says derivkit API
- move it perhaps to the bottom in the inde so it is the last thing on the menu

docs/api/modules.rst
nikosarcevic: can we not have toc level 4. i think level 1 or 2 might be enough. level 1 woudl then only show kits which is the best way for a user ot see rather than expand this inot level 4 WDYT

nikosarcevic:
@lonbar 

we may have a solution for the url mess

add this to extensions --     "sphinxext.rediraffe",


```
extensions = [
    "sphinx.ext.autodoc",
    "sphinx.ext.napoleon",
    "sphinx.ext.githubpages",
    "sphinx.ext.intersphinx",
    "sphinx.ext.doctest",
    "matplotlib.sphinxext.plot_directive",
    "sphinx_design",
    "sphinx_multiversion",
    "sphinx_copybutton",
    "adoption",
    "sphinxext.rediraffe",
]
```
 
and this to conf.py

```
# -----------------------------------------------------------------------------
# Redirects (keep old API URLs working after moving generated pages to /api/)
# -----------------------------------------------------------------------------
# Requires: sphinxext-rediraffe
# This creates redirect stub pages so old links like:
#   /derivkit.calculus_kit.html
# now redirect to:
#   /api/derivkit.calculus_kit.html
api_dir = Path(__file__).parent / "api"

rediraffe_redirects = {}
if api_dir.is_dir():
    for rst in api_dir.glob("derivkit*.rst"):
        stem = rst.stem  # e.g. "derivkit.calculus_kit"
        rediraffe_redirects[stem] = f"api/{stem}"
```

lonbar:
> I am not sure i ujderstand the point 2 -- you are sayign that once we move stuff into api dir, the urls are gonna be ugly liek you showed?

I mean that if the original URL is, say, https://example.com/link, it would change to https://example.com/api/link after this merge. It would only affect the API reference pages.

lonbar:
Well, if the original URL is long and weird, the new URL would still be long and weird. It would just have an additional `/api/` attached.

PR #400, opened 2026-01-31
Propagate DALI triplet to ForecastKit

### LINKED ISSUE #342: DALI API should expose expansion order
Users care about the DALI expansion order, not “doublet vs triplet”. Update the API so users can request the expansion up to a given order.

Proposal
fk.dali(expansion_order=2)  # doublet, this should be default
fk.dali(expansion_order=3)  # triplet

NOTE: we should NOT call this arg just `order` since we do use order in derivatives so i tihnk we should not duplicate args.

Tasks
Add expansion_order (or max_order) argument to ForecastKit.dali
Explain clearly in the docstring what each order will return aka F, G or F, G, H or F, G, H, I....
Update docs + tests

RELATED NEW ISSUE:
once this is in and merged just expand getdist dali samples (and related methods) to be able to handle triplet sampling!!

@lonbar you can decide if u wanna keep it as a child issue or part of this or a separate issue.

lonbar:
I have updated `build_dali()` and `ForecastKit.dali()` to use the new functionality of `get_forecast_tensors()`. A user is now able to make use of the DALI triplet.

Closes #342.

### DIFF
--- a/src/derivkit/forecasting/dali.py
+++ b/src/derivkit/forecasting/dali.py
@@ -78,9 +78,10 @@ def build_dali(
         would be ``DALI_triplet``. In the case that ``forecast_order == 1``
         the Fisher matrix itself is returned instead of ``DALI_singlet``.
 
-        The number of axes of the tensors in the multiplet at order ``order``
-        ranges from ``order+1`` to ``2*order`` in order of increasing number.
-        All axes have length ``len(theta0)``.
+        The tensors in the multiplet at order ``O`` have parameter axes that
+        increase in number from ``O+1`` to ``2*O`` inclusive. For example, the
+        DALI doublet tensors have 3 and 4 axes, respectively. All axes have
+        length ``len(theta0)``.
     """
     return get_forecast_tensors(
         function,


src/derivkit/forecasting/dali.py
```diff
@@ -40,22 +46,45 @@ def build_dali(
...
         n_workers: Number of workers for per-parameter parallelization/threads.
             Default ``1`` (serial). Inner batch evaluation is kept serial to
             avoid oversubscription.
         dk_kwargs: Additional keyword arguments passed to
             :class:`derivkit.calculus_kit.CalculusKit`.
 
     Returns:
-        A tuple ``(G, H)`` where ``G`` has shape ``(P, P, P)`` and ``H`` has
-        shape ``(P, P, P, P)``, with ``P`` being the number of model parameters.
+        If ``single_forecast_order`` is ``False`` the result is a dictionary
```
nikosarcevic: can you keep the shape explanation as well as it is very informative aka Fisher is (p, p), doublet dali first tensor has shape (p, p, p) etc. I think we keep lowecase p but please confirm this is the case for our convention)

src/derivkit/forecast_kit.py
```diff
@@ -272,25 +274,52 @@ def dali(
...
+                    3: DALI_triplet,
+                }
+
+            where ``DALI_singlet`` contains the Fisher matrix, ``DALI_doublet``
+            contains the DALI doublet tensors, etc.
+
+            If ``single_forecast_order`` is ``True`` the result instead is the
+            multiplet of the requested order. Using the example above the output
+            would be ``DALI_triplet``. In the case that ``forecast_order == 1``
+            the Fisher matrix itself is returned instead of ``DALI_singlet``.
```
nikosarcevic: same as above -- I think it is great to keep explicit doublet returns two tensors D1 and D2 with shapes ppp and pppp respectively etc

src/derivkit/forecasting/dali.py
```diff
@@ -40,22 +46,49 @@ def build_dali(
...
+        contains the DALI doublet tensors, etc.
+
+        If ``single_forecast_order`` is ``True`` the result instead is the
+        multiplet of the requested order. Using the example above the output
+        would be ``DALI_triplet``. In the case that ``forecast_order == 1``
+        the Fisher matrix itself is returned instead of ``DALI_singlet``.
+
+        The number of axes of the tensors in the multiplet at order ``order``
+        ranges from ``order+1`` to ``2*order`` in order of increasing number.
+        All axes have length ``len(theta0)``.
```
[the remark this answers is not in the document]
lonbar: Included a rephrased version of your second suggestion.

PR #401, opened 2026-01-31
Fix/lint unsafe fixes

nikosarcevic:
in the PR I have fixed mostly lintign problems in forecast_kit, stencil and dali scripts

i have also fixed unused vars in stencil and improved code a bit

locally linting healthy, pytests pass, docs are built

### DIFF
--- a/src/derivkit/forecasting/dali.py
+++ b/src/derivkit/forecasting/dali.py
@@ -2,16 +2,12 @@
 
 from __future__ import annotations
 
-from typing import Any, Callable, Union
+from typing import Any, Callable
 
 import numpy as np
-from numpy.typing import NDArray
 
 from derivkit.forecasting.forecast_core import get_forecast_tensors
-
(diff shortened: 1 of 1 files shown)


src/derivkit/derivatives/finite/stencil.py
```diff
@@ -5,7 +5,9 @@
 import numpy as np
 from numpy.typing import NDArray
 
-__all = [
+FloatArray = NDArray[np.float64]
```
lonbar: It seems more convenient to define a separate module for all the types (`derivkit.types`), as otherwise it could quickly become unclear what each type represents.

lonbar (commented):
Thanks for catching this. I'm a bit annoyed that our tox pipeline didn't catch this. It's probably not configured correctly. I'll have a look into this.

lonbar (approved):

PR #417, opened 2026-02-06
vector-valued hessian fixes

ctrendafilova:
This is functional fixes only. Docstrings and tests still need to be updated.

### DIFF
--- a/src/derivkit/calculus/hessian.py
+++ b/src/derivkit/calculus/hessian.py
@@ -151,12 +151,8 @@ def _build_hessian_full(
         inner_workers=inner_workers,
     )
     y0 = np.asarray(function(theta))
-    if y0.ndim == 0:
-        hess = np.empty((1, p, p), dtype=float)
-    else:
-        out_shape = y0.shape
-        o = int(out_shape[0])
-        hess = np.empty((o, p, p), dtype=float)
+    out_shape = y0.shape
+    hess = np.empty((*out_shape, p, p), dtype=float)
     k = 0
     vals = np.array(vals_list, dtype=float)
     for i in range(p):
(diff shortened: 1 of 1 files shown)


ctrendafilova:
Docstrings have been updated. I will fix tests in the next commits.

src/derivkit/calculus/hessian.py
```diff
@@ -183,11 +121,10 @@ def _build_hessian_scalar_full(
             :meth:`derivkit.derivative_kit.DerivativeKit.differentiate`.
 
     Returns:
-        A 2D array representing the Hessian.
+        An ND array representing the Hessian.
 
     Raises:
         FloatingPointError: If non-finite values are encountered.
-        TypeError: If ``function`` does not return a scalar value.
```
lonbar: Is `_build_hessian_full()` able to differentiate functions of an arbitrary dimension? Say, a function returning an array of shape `(1, 2, 4)`? If so, the summary line isn't correct, because it states it only handles scalar- and vector-valued functions. If it doesn't, it should probably `raise` an exception.

src/derivkit/calculus/hessian.py
```diff
@@ -200,39 +137,48 @@ def _build_hessian_scalar_full(
...
+            hess[:,i, j] = hij
+            hess[:,j, i] = hij
 
     if not np.isfinite(hess).all():
         raise FloatingPointError("Non-finite values encountered in Hessian.")
-    return hess
+    if y0.ndim == 0:
+        return hess[0,:,:]
+    else:
+        return hess
```
lonbar:
```suggestion
        hess[..., i, i] = vals[k]
        k += 1
    for i in range(p):
        for j in range(i + 1, p):
            hij = vals[k]
            k += 1
            hess[..., i, j] = hij
            hess[..., j, i] = hij

if not np.isfinite(hess).all():
    raise FloatingPointError("Non-finite values encountered in Hessian.")

return hess
```
The ellipsis expands to however many `:`s are required. See [dimensional indexing tools](https://numpy.org/doc/stable/user/basics.indexing.html#dimensional-indexing-tools).

PR #418, opened 2026-02-06
Feature/vectorized derivatives cache

### LINKED ISSUE #416: Excessive model evaluations in Hessian / higher-order derivatives for vector outputs
When computing Hessians (and higher-order DALI derivatives) for **vector-valued forward models**, the current implementation triggers a very large number of repeated model evaluations. This becomes a major performance bottleneck for expensive models (e.g. PyCCL Cls).

### Cause

Two factors drive the overhead:

1. **Per-component scalarization**
   Vector outputs are flattened and differentiated one scalar component at a time, so the full model is recomputed for each output entry and stencil point.

2. **Nested mixed partials**
   Mixed Hessian terms are computed as derivatives of derivatives, multiplying stencil evaluations instead of using direct mixed-partial stencils.

### Effect

* Very high derivative request counts
* Many repeated evaluations at identical θ
* Simple caching drastically reduces runtime

So this is a performance/architecture issue, not a correctness bug

nikosarcevic:
this PR avoids repeated calls to the forward model when building hessians, hyperhessians, and forecast tensors. 

The model is evalutaed once at the refernece params to check shapes and covariance consitsency and that result is reused instead of recomputing it over nad over

a cached model wrapper is then passed through the derivative code so repeated evaluations at the same parameter values return the cached result. This significantly reduces runtime for expensive models without changing any numerical results

now this is not a perfect solution but it is a great intermeidate solution without chaging much in the code -- in order to rto do it properly, we would have to break the derivation kit api and make the kit accept solar/vec/tensor INPUT AND OUTPUT (currently only input works)

but we do not want to do that now so this is a good intermeidate solution 

i have tested the new code on camb and ccl and it helps tremendously 

this closes #416

### DIFF
--- a/tests/test_forecast_forecast_core.py
+++ b/tests/test_forecast_forecast_core.py
@@ -380,7 +380,9 @@ def test_scalar_dali_triplet(model, theta, expected):
         if np.all(exp == 0):
             assert np.allclose(triplet[i], exp, atol=2e-5, rtol=0.0)
         else:
-            assert np.allclose(triplet[i], exp, atol=1e-8, rtol=1e-7)
+            exp_abs = float(np.max(np.abs(exp)))
+            atol = max(1e-8, 1e-4 if exp_abs > 1 else 2e-5)
+            assert np.allclose(triplet[i], exp, atol=atol, rtol=3e-7)
 
 
 def get_all_indices(max_indices):


nikosarcevic:
@lonbar is working on caching only (issue #427 ) so I will close this PR.

PR #430, opened 2026-02-08
Add function caching module

### LINKED ISSUE #427: Runtime of forecasts can be improved
Running a forecast for a reasonably realistic model takes a much longer time than expected.
@ctrendafilova improved the runtime of higher-order derivatives in #417, which keeps the runtime managable.
I have run a cosmic shear forecast up to the DALI doublet for a weak-lensing model containing 3 parameters ($\Omega_m$, $\sigma_8$ and the IA amplitude) with CCL.
The model includes 5 redshift bins (15 correlations) and 60 different $C_\ell$ s.
Running the entire model takes around 6 minutes on a single core.

Further improvements can be made by caching the derivatives. This allows DerivKit to look up previously computed values, which means it doesn't need to recompute the same derivative again.

Because of DerivKit's architecture it isn't immediately obvious how caching should be implemented, but a reasonable first step is to add a caching module that construct a lookup table. The question is then how this impacts resource usage and runtime. I want to use this thread to record some results of benchmarks I am currently running.

lonbar:
This adds a small value caching to `ForecastKit`, which speeds up the forecast considerable (a quick test is sped up by about 63% on my laptop).

The parameters of the caching can be tweaked by passing values to `ForecastKit`. By default the caching is turned off.

Co-authored by @nikosarcevic.

Closes #427.

### DIFF
--- a/src/derivkit/forecast_kit.py
+++ b/src/derivkit/forecast_kit.py
@@ -101,7 +101,7 @@ class ForecastKit:
                  | Callable[[np.ndarray], np.ndarray],
             *,
             cache_theta: bool = True,
-            cache_theta_number_decimals: int = 14,
+            cache_theta_number_decimal_places: int = 14,
             cache_theta_maxsize: int | None = 4096,
     ):
         r"""Initialises the ForecastKit with model, fiducials, and covariance.
@@ -129,14 +129,14 @@ class ForecastKit:
 
             cache_theta: A flag which, if set to ``True``, turns on caching of
                 function values.
-            cache_theta_number_decimals:  The number of decimals that are
+            cache_theta_number_decimal_places:  The number of decimal places that are
                 included in the caching.
             cache_theta_maxsize: The maximum size of the cache.
         """
         if cache_theta:
             function = wrap_theta_cache_builtin(
                 function,
-                number_decimals=cache_theta_number_decimals,
+                number_decimal_places=cache_theta_number_decimal_places,
                 maxsize=cache_theta_maxsize,
                 copy=True,
             )
--- a/src/derivkit/utils/caching.py
+++ b/src/derivkit/utils/caching.py
@@ -12,18 +12,18 @@ import numpy as np
 def wrap_theta_cache_builtin(
     function: Callable[[np.ndarray], np.ndarray],
     *,
-    number_decimals: int = 14,
+    number_decimal_places: int = 14,
     maxsize: int | None = 4096,
     copy: bool = True,
 ) -> tuple[Callable[[np.ndarray], np.ndarray], dict[str, Any]]:
     """Creates a cache for function values.
 
     As part of the caching the input is truncated to a pre-set number
-    of decimal digits.
+    of decimal places.
 
     Args:
         function: The function to be cached.
-        number_decimals: The number of decimals to include before
+        number_decimal_places: The number of decimal places to include before
             truncating.
         maxsize: The size of the cache.
         copy: A flag that, when set to ``True``, causes the function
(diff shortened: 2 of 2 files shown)


src/derivkit/utils/caching.py
```diff
@@ -0,0 +1,59 @@
...
+from functools import lru_cache, wraps
+from typing import Any
+
+import numpy as np
+
+
+def wrap_theta_cache_builtin(
+    function: Callable[[np.ndarray], np.ndarray],
+    *,
+    number_decimals: int = 14,
```
nikosarcevic: i tihnk elsewhere we have decimal_places as an arg

src/derivkit/utils/caching.py
```diff
@@ -0,0 +1,59 @@
...
+            to create a copy of the function input.
+
+    Returns:
+        A caching wrapper of :func:`function` if it is a ``Callable``.
+        If :func:`function: is not a ``Callable`` it will be returned instead.
+    """
+    @lru_cache(maxsize=maxsize)
+    def cached_wrapper(
+        cachable_array: tuple[float, ...]
+    ) -> np.ndarray[float, ...]:
```
nikosarcevic: add a one liner docstring

src/derivkit/forecast_kit.py
```diff
@@ -98,9 +99,18 @@ def __init__(
             theta0: Sequence[float] | np.ndarray,
             cov: np.ndarray
                  | Callable[[np.ndarray], np.ndarray],
+            *,
+            cache_theta: bool = False,
```
nikosarcevic: absolutely set to True.

src/derivkit/utils/caching.py
lonbar: I have used `grep` to search through the code, but haven't found it.

src/derivkit/forecast_kit.py
lonbar: Changed this in 997ac26df98f1378ac93968210610900f065f53b (I typed "off", but I promise the caching is turned on!).

src/derivkit/utils/caching.py
```diff
@@ -0,0 +1,59 @@
...
+def wrap_theta_cache_builtin(
+    function: Callable[[np.ndarray], np.ndarray],
+    *,
+    number_decimals: int = 14,
+    maxsize: int | None = 4096,
+    copy: bool = True,
+) -> tuple[Callable[[np.ndarray], np.ndarray], dict[str, Any]]:
+    """Creates a cache for function values.
+
+    As part of the caching the input is truncated to a pre-set numbber
```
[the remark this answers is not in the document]
lonbar: I compromised and wrote "decimal digits" :slightly_smiling_face:.

src/derivkit/utils/caching.py
```diff
@@ -0,0 +1,59 @@
+"""Provides :func:`wrap_theta_cache_builtin`."""
```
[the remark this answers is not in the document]
lonbar: Added in 9c84533f23a6af3bf3e04bcfb26eaa055f59c608.

nikosarcevic:
@lonbar this should also close #416 I think

lonbar:
@nikosarcevic Does it? I thought the remaining issue was due to the implementation of the (hyper-)hessian.

nikosarcevic (approved):
tysm

nikosarcevic:
> @nikosarcevic Does it? I thought the remaining issue was due to the implementation of the (hyper-)hessian.

i tihnk so. i tihnk we should either reword that issue or write a new one

nikosarcevic:
> @nikosarcevic Does it? I thought the remaining issue was due to the implementation of the (hyper-)hessian.

i tihnk so. i tihnk we should either reword that issue or write a new one

PR #432, opened 2026-02-08
Add v1.1.0 entry for caching feature

lonbar:


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## v1.1.0
+
+### Added
+
+* Added caching of function and derivative values for forecasting.
```
nikosarcevic: do we need to specify it is in forecast kit only?

PR #439, opened 2026-02-10
Fix/reexport kits 437

### LINKED ISSUE #437: Re-export all kits at top-level (from derivkit import ...)
Users currently have to import kits one layer deep (e.g. from derivkit.forecast_kit import ForecastKit). This is not a clean public API.

Fix
Re-export all user-facing kits from derivkit/__init__.py so users can do:
from derivkit import ForecastKit, CalculusKit, ...

nikosarcevic:
Adds the following:

* Improves the usage examples of the kits
* Updated module atrtibute of the kits
* added unit tests to test the public API

Closes #437.

### DIFF
--- a/tests/test_public_api.py
+++ b/tests/test_public_api.py
@@ -7,7 +7,9 @@ import derivkit
 
 def test_public_all_contains_kits():
     """Test that __all__ contains the expected public kits."""
-    assert set(derivkit.SUPPORTED_KITS).issubset(set(derivkit.__all__))
+    public = set(derivkit.__all__)
+    for cls in derivkit.SUPPORTED_KITS:
+        assert cls.__name__ in public
 
 
 def test_kits_report_top_level_module():


tests/test_public_api.py
```diff
@@ -0,0 +1,27 @@
...
+    assert ForecastKit is not None
+    assert DerivativeKit is not None
+    assert CalculusKit is not None
+    assert LikelihoodKit is not None
+
+
+def test_public_all_contains_kits():
+    """Test that __all__ contains the expected public kits."""
+    expected = {"CalculusKit", "DerivativeKit", "ForecastKit", "LikelihoodKit"}
+    assert expected.issubset(set(derivkit.__all__))
```
lonbar: Is this useful? It would have to be updated with each new kit.

tests/test_public_api.py
```diff
@@ -0,0 +1,17 @@
+"""Unit tests for public API."""
+
+from __future__ import annotations
+
+import derivkit
+
+
+def test_public_all_contains_kits():
+    """Test that __all__ contains the expected public kits."""
+    expected = {"CalculusKit", "DerivativeKit", "ForecastKit", "LikelihoodKit"}
```
lonbar:
```suggestion
```

PR #447, opened 2026-02-13
Concurrency: thread locking option + spawn-based process execution with safe env caps

### LINKED ISSUE #446: Default to process-based parallelism; avoid threads for native-threaded models (CCL/CAMB)
When DerivKit is used with models that repeatedly call `pyccl.angular_cl` (cosmic shear / 3x2pt / other probes), setting `n_workers > 1` with the current thread-based executor can hang or segfault (CCL/CAMB/OpenMP/BLAS interaction). Instability can also occur with `n_workers=1` unless native threadpools are clamped to 1.

Observed:

* `n_workers > 1` (threads) → concurrent `angular_cl` in one process → hang/segfault.
* `n_workers = 1` can still be unstable if `OMP_NUM_THREADS` / BLAS threads > 1.
* Stable only when setting before imports:

  ```
  OMP_NUM_THREADS=1
  OPENBLAS_NUM_THREADS=1
  MKL_NUM_THREADS=1
  VECLIB_MAXIMUM_THREADS=1
  NUMEXPR_NUM_THREADS=1
  ```

**Proposal fix:**

* Make **process-based** outer parallelism the default (or auto-switch when `n_workers>1`), and avoid `ThreadPoolExecutor` for model evals.
* Enforce a **single level of parallelism** for nested derivatives: if outer workers >1, set `inner_workers=1` by default.
* Optional: provide a native-thread clamp utility (e.g. `threadpoolctl`) and document “threads=1 per process” best practice.

Repro notebook: [link](https://github.com/derivkit/derivkit-demos/blob/main/demo-notebooks/10-derivkit-ccl-demo.ipynb)

nikosarcevic:
This PR adds an opt-in thread-safety wrapper for model callables and improves concurrency handling to reduce nested-parallel oversubscription.

### Changes

* Added `thread_safe` / `thread_lock` to `ForecastKit` and `CalculusKit` to optionally serialize model evaluation.
* Added `backend="processes"` to `parallel_execute` (spawn-based) with safe propagation of inner worker settings.
* For process backend, child processes default to `OMP/MKL/OPENBLAS/... = 1` to avoid thread oversubscription (especially on macOS).
* Clearer error message if a non-picklable worker is used with the process backend.
* Existing outer/inner worker resolution logic preserved.

### Compatibility

* Fully backward compatible.
* Locking is opt-in and off by default.
* Process backend is additive.

### Tests

* Added tests covering locking behavior and `thread_lock` forwarding.

pytest passes, linting healthy, docs build locally

this would close #446

### DIFF
--- a/src/derivkit/utils/concurrency.py
+++ b/src/derivkit/utils/concurrency.py
@@ -137,12 +137,12 @@ def _run_in_child(
     """
     worker, args, inner_workers, env = payload
 
-    # If backend="processes", this function runs inside each spawned worker process.
-    # Any provided environment variables (e.g. OMP_NUM_THREADS,
-    # OPENBLAS_NUM_THREADS, MKL_NUM_THREADS) are set here so that
-    # thread-pool sizes are explicitly fixed per process before the
-    # worker (and any heavy native imports it may trigger) is executed.
-    # Users may override these defaults by passing custom values.
+    # If backend="processes", this runs inside each spawned worker process.
+    # Set environment variables here (inside the child) *before* running the worker
+    # (and before any heavy native imports it may trigger). This is primarily used to
+    # control native thread pools (OpenMP/BLAS/vecLib/NumExpr) per process.
+    # By default, parallel_execute sets common thread-count variables to "1" in each child
+    # (see _default_child_env), unless the user overrides them via child_env.
     if env:
         for k, v in env.items():
             os.environ[k] = str(v)


src/derivkit/utils/concurrency.py
```diff
@@ -93,6 +94,37 @@ def _detect_hw_threads() -> int:
...
+        "MKL_NUM_THREADS": "1",
+        "VECLIB_MAXIMUM_THREADS": "1",
+        "NUMEXPR_NUM_THREADS": "1",
+    }
+
+
+def _run_in_child(
+    payload: tuple[Callable[..., Any],
+    tuple[Any, ...], int | None,
+    dict[str, str] | None]
```
lonbar:
```suggestion
    payload: tuple[
        Callable[..., Any],
        tuple[Any, ...],
        int | None,
        dict[str, str] | None
    ]
```
I was a bit confused about what the input represents, so perhaps this clarifies that we're dealing with a single input parameter, which is a tupple of a function, a tuple, an int and a dict.

src/derivkit/utils/concurrency.py
```diff
@@ -93,6 +94,37 @@ def _detect_hw_threads() -> int:
...
+) -> Any:
+    worker, args, inner_workers, env = payload
```
lonbar:
```suggestion
    """Executes a function in a child process.

    Args:
        payload: the elements of the child process to run.
            Is a tuple with the elements ``(f, args, pid, env)`` where
            
            * ``f`` is the function to evaluate.
            * ``args`` is a tuple of arguments to which ``f`` is applied.
            * ``pid`` is a token identifying the child process.
              Note that ``pid`` is internal and not exposed to the operating system.
            * ``env`` is a dictionary containing variables names with their values.
    """
    worker, args, inner_workers, env = payload
```

src/derivkit/utils/concurrency.py
```diff
@@ -93,6 +94,37 @@ def _detect_hw_threads() -> int:
...
+
+    # Clamp threads INSIDE the child process, before heavy imports happen there.
+    if env:
+        for k, v in env.items():
+            os.environ[k] = str(v)
+
+    # Propagate inner-workers setting in the child too
```
lonbar: This is where the child is executed, but it isn't immediately clear to me from this function alone where multiprocessing comes in. Perhaps we can add a comment here?

src/derivkit/utils/thread_safety.py
```diff
@@ -0,0 +1,25 @@
+"""Thread safety utilities."""
+
+from __future__ import annotations
+
+import threading
+from collections.abc import Callable
+from typing import Any, TypeVar
+
+T = TypeVar("T")
```
lonbar: Is this in pratice different from `Any`?

tests/test_utils_concurrency.py
```diff
@@ -206,12 +217,9 @@ def worker(x): return x + 1
 
 def test_parallel_execute_backend_processes_not_implemented():
     """Tests that backend='processes' raises NotImplementedError."""
```
lonbar: This is no longer tested.

tests/test_utils_concurrency.py
```diff
@@ -206,12 +217,9 @@ def worker(x): return x + 1
...
-    def identity(x: int) -> int:
-        """Identity function used for testing."""
-        return x
-
-    with pytest.raises(NotImplementedError):
-        parallel_execute(identity, [(1,)], backend="processes")
+    out = parallel_execute(_identity, [(1,), (2,)], outer_workers=2,
+                           backend="processes", child_env={})
+    assert sorted(out) == [1, 2]
```
lonbar: Does this check parallelisation/serialisation?

src/derivkit/utils/concurrency.py
nikosarcevic: Yep — I’ll add a short note that _run_in_child is the function invoked via ProcessPoolExecutor.map(...) in the backned="processes" path, and clarify  (it carries inner_workers, not a PID)

82293cb2d78a263b762f5310dca92b62331fd573

src/derivkit/utils/concurrency.py
```diff
@@ -131,29 +164,51 @@ def parallel_execute(
...
                 futures = []
                 for args in arg_tuples:
-                    # Each task gets its own copy of the current context
                     ctx = contextvars.copy_context()
                     futures.append(ex.submit(ctx.run, worker, *args))
                 return [f.result() for f in futures]
-        else:
-            return [worker(*args) for args in arg_tuples]
+
+        # processes (spawn)
```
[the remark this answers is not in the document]
nikosarcevic: clarified in ab2ce49059058e8158c3ddecd26fea8f85bce009

tests/test_utils_concurrency.py
```diff
@@ -233,7 +241,7 @@ def worker(_x: int) -> int:
 
     tasks = [(i,) for i in range(n)]
 
-    out = parallel_execute(worker, tasks, outer_workers=n, inner_workers=1)
+    out = parallel_execute(worker, tasks, outer_workers=n, inner_workers=1, backend="threads")
```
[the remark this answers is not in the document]
nikosarcevic: we can keep it as is i guess

src/derivkit/utils/concurrency.py
lonbar: I've added some comments to that commit. This still leaves the question of whether this comment by itself is useful. Is it helpful to extend it so that it explains that "inner-workers setting" means that the child processes allow the same number of `inner_workers`?

Issue #461, opened 2026-04-11
Polyfit/adaptive derivative backends can suppress the non-Gaussian structure needed by DALI


nikosarcevic:
When computing higher-order derivatives for DALI (forecast_order ≥ 2), smoothing-based derivative backends such as polyfit and adaptive wash out the local non-Gaussian structure that DALI is meant to capture.

I have tried playing with polyfit and adaptive kwargs, but tuning them does not fix the problem reliably. This is not just a tuning issue: these methods are built to smooth and stabilize, which works against resolving higher-order local curvature.

At this point, higher-order DALI derivatives need to fall back internally to finite differences with Ridders extrapolation. For Fisher / first-order derivatives, the smoother backends are fine.

nikosarcevic:
resolved with #462

PR #462, opened 2026-04-11
DALI deriv backend fix 

nikosarcevic:
in this PR I have:

- expanded dali tests to have more checks for deriv backends
- added a backend policy into forecast core when computing dali tensors to fallback on finite with ridders so we do not wash out the nongaussianities
- added a small ccl consistency check (real-ish analysis) to confirm the backend policy works

this solves #461

### DIFF
--- a/src/derivkit/forecasting/forecast_core.py
+++ b/src/derivkit/forecasting/forecast_core.py
@@ -20,6 +20,7 @@ from numpy.typing import NDArray
 from derivkit.calculus_kit import CalculusKit
 from derivkit.utils.concurrency import normalize_workers
 from derivkit.utils.linalg import invert_covariance
+from derivkit.utils.logger import derivkit_logger
 from derivkit.utils.types import ArrayLike1D, ArrayLike2D
 from derivkit.utils.validate import validate_covariance_matrix_shape
 
@@ -29,11 +30,11 @@ __all__ = [
 ]
 
 
-#: The supported orders of the DALI expansion.
-#:
-#: A value of 1 corresponds to the Fisher matrix.
-#: A value of 2 corresponds to the DALI doublet.
-#: A value of 3 corresponds to the DALI triplet.
+#  The supported orders of the DALI expansion.
+#
+#  A value of 1 corresponds to the Fisher matrix.
+#  A value of 2 corresponds to the DALI doublet.
+#  A value of 3 corresponds to the DALI triplet.
 SUPPORTED_FORECAST_ORDERS = (1, 2, 3)
 
 SUPPORTED_DERIVATIVE_ORDERS = (1, 2, 3)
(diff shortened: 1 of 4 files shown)


src/derivkit/forecasting/forecast_core.py
```diff
@@ -29,11 +30,11 @@
...
-#: The supported orders of the DALI expansion.
-#:
-#: A value of 1 corresponds to the Fisher matrix.
-#: A value of 2 corresponds to the DALI doublet.
-#: A value of 3 corresponds to the DALI triplet.
+#  The supported orders of the DALI expansion.
+#
+#  A value of 1 corresponds to the Fisher matrix.
+#  A value of 2 corresponds to the DALI doublet.
+#  A value of 3 corresponds to the DALI triplet.
```
lonbar:
```suggestion
#:  The supported orders of the DALI expansion.
#:
#:  A value of 1 corresponds to the Fisher matrix.
#:  A value of 2 corresponds to the DALI doublet.
#:  A value of 3 corresponds to the DALI triplet.
```
[Sphinx requires the `:`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings) as `SUPPORTED_FORECAST_ORDERS` does not have a `__doc__` attribute. Without the `:`, `SUPPORTED_FORECAST_ORDERS` would not be documented.

tests/benchmarks/test_dali_backend_consistency.py
```diff
@@ -0,0 +1,292 @@
+"""Small fast backend check for Fisher + DALI(2) on a reduced CCL problem.
```
lonbar: (2) refers to the doublet, right?

tests/test_forecast_forecast_core.py
lonbar, line 1: Is this relevant? I mean, it's a factor of 5, but both numbers are very small.

tests/test_forecast_kit_dali.py
```diff
@@ -311,3 +311,310 @@ def test_build_dali_forwards_arbitrary_extra_kwargs(dali_mocks):
...
+    "method,build_kwargs",
+    [
+        ("finite", {"extrapolation": None, "num_points": 5, "stepsize": 1e-4}),
+        ("finite", {"extrapolation": "ridders", "num_points": 5, "stepsize": 1e-3}),
+        ("finite", {"extrapolation": "richardson", "num_points": 5, "stepsize": 1e-3}),
+        ("polyfit", {"degree": 4}),
+        ("adaptive", {"n_points": 10, "spacing": "auto"}),
+    ],
+)
+def test_build_dali_preserves_expected_tensor_symmetries_for_real_methods(
```
lonbar, line 537: Isn't this a test for the DALI implementation proper?

tests/test_forecast_kit_dali.py
```diff
@@ -311,3 +311,310 @@ def test_build_dali_forwards_arbitrary_extra_kwargs(dali_mocks):
...
+    "method,build_kwargs",
+    [
+        ("finite", {"extrapolation": None, "num_points": 5, "stepsize": 1e-4}),
+        ("finite", {"extrapolation": "ridders", "num_points": 5, "stepsize": 1e-3}),
+        ("finite", {"extrapolation": "richardson", "num_points": 5, "stepsize": 1e-3}),
+        ("polyfit", {"degree": 4}),
+        ("adaptive", {"n_points": 10, "spacing": "auto"}),
+    ],
+)
+def test_build_dali_returns_small_tensors_for_linear_model_across_methods(
```
lonbar, line 579: Isn't this too a test for the DALI proper implementation?

tests/benchmarks/test_dali_backend_consistency.py
lonbar, line 1: This isn't meant to be run by the CI, right?

lonbar (commented):
Thanks for catching and fixing this! Added a few minor comments.

tests/test_forecast_forecast_core.py
nikosarcevic: it is relevant as we changed the deriv backend whihc is not doing the smoothing liek poly and adaptive are so is is "less accurate". this passed tests locally on mac but linux ci failed so i had to slightly change the tolerance

tests/benchmarks/test_dali_backend_consistency.py
nikosarcevic: fixed 55a968859b5d544f0ecdee7db8b093b64e8cee17

tests/test_forecast_kit_dali.py
nikosarcevic: yes on a toy model, not ccl.

tests/test_forecast_kit_dali.py
nikosarcevic: yes, just another test to cover the basis (parametrize)

tests/benchmarks/test_dali_backend_consistency.py
nikosarcevic: no this is marked as slow and will be run only when we wnt to (I most certainly hope so)

lonbar (approved):

PR #464, opened 2026-04-12
Fix broken docs build on release tags

nikosarcevic:
Use tox to manage docs dependencies instead of relying on pip group installs.
Removes brittle setup and ensures consistent Sphinx builds in CI.


lonbar (approved):
МРЏ

PR #466, opened 2026-05-28
added gh icon in the docs footer

### LINKED ISSUE #465: Link to GitHub project from docs
We currently do not have a direct GH link to the repo from the docs

the best way is to just do it like we did in binny: [see this PR in binny](https://github.com/binny-org/binny/pull/65).

This is a trivial change.

nikosarcevic:
in this pr i slightly modified conf in docs so that the gh link to the derivkit repo is in the docs footer 

closes #465

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -174,7 +174,7 @@ if html_theme == "furo":
         "footer_icons": [
             {
                 "name": "GitHub",
-                "url": "https://github.com/derivkit-org/derivkit",
+                "url": "https://github.com/derivkit/derivkit",
                 "html": """
                 <svg stroke="var(--color-foreground-primary)"
                      fill="var(--color-foreground-primary)"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -167,7 +167,6 @@ html_permalinks_icon = "<span>#</span>"
 if html_theme == "furo":
     html_theme_options = {
         "source_repository": "https://github.com/derivkit-org/derivkit/",
-        "source_branch": "main",
         "source_directory": "docs/",
         "light_logo": "logos/logo-black.png",
         "dark_logo": "logos/logo-blue.png",
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -166,11 +166,36 @@ html_permalinks_icon = "<span>#</span>"
 
 if html_theme == "furo":
     html_theme_options = {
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -166,11 +166,36 @@
...
         "light_logo": "logos/logo-black.png",
-        "dark_logo": "logos/logo-blue.png",  # or logo-black.png if you prefer
-        # keep the project name visible (logo + text)
-        # "sidebar_hide_name": True,  # <- remove this
-
+        "dark_logo": "logos/logo-blue.png",
+        "footer_icons": [
+            {
+                "name": "GitHub",
+                "url": "https://github.com/derivkit-org/derivkit",
```
lonbar:
```suggestion
                "url": "https://github.com/derivkit/derivkit",
```

docs/conf.py
```diff
@@ -166,11 +166,36 @@
 
 if html_theme == "furo":
     html_theme_options = {
+        "source_repository": "https://github.com/derivkit-org/derivkit/",
+        "source_branch": "main",
```
lonbar: From a quick glance I was unable to find what this does. Does this play nicely with sphinx-multiversion? The part of the documentation that a user is looking at might not belong to the main branch.

lonbar (changes requested):
I like it. Thanks for adding this. I noticed that the organisation was misspelt, so I'm suggesting a change.

The `source_branch` option is a bit strange to me. The user might not look at the documenation of the main branch, so this is potentially misleading.

docs/conf.py
nikosarcevic: ah good point. These Furo options add the “Edit/View source” links to the theme. Since we build versioned docs with sphinx-multiversion, hardcoding `main` is not ideal i guess because a user viewing a tagged version could be sent to the source on `main`.

I’ll either remove these source-link options for now, or make the branch/tag dynamic if we decide we want them. what do yo uthink?

lonbar: removing this line gives a broken URL though. In this case the value seems to be set to `''`, which shows up in the URL as a repeated `/`.

I'm not really sure what's the best thing to do at the moment.

lonbar: As we've discussed via call we can leave this line out.

lonbar:
```suggestion
```

lonbar (approved):

PR #471, opened 2026-06-28
Fix tox CI environment selection

nikosarcevic:
Fixes CI by making each Python matrix job run only the matching tox environment.

Previously, `uv run tox -m test` ran all labeled test environments (`py310`, `py311`, `py312`, `py313`) inside every matrix job. Since each job only installs one Python version, tox failed when the other interpreters were missing.

This PR also makes the docs environment explicit by adding missing runtime dependencies needed by the Sphinx theme/extensions. This avoids relying on transitive dependencies that may change upstream.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         run: uv sync --group test
 
       - name: Run tests
-        run: uv run tox -m test
+        run: uv run tox -e py
 
   linting:
     runs-on: ${{ matrix.os }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,9 @@ docs = [
   "sphinx-multiversion",
   "furo",
   "sphinx-copybutton",
+  "six",
+  "typing_extensions",
+  "charset_normalizer",
   {include-group = "dev"},
 ]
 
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: uv sync --group test
 
       - name: Run tests
-        run: uv run tox -m test
+        run: uv run tox -e py
```
lonbar, line 32: It might be useful to update this in the docs, as we currently instruct people to run the `test` dependency group. This is fine in principle, but because a recent change in tox this will cause errors for versions of python which are not available.

pyproject.toml
```diff
@@ -54,6 +54,9 @@ docs = [
   "sphinx-multiversion",
   "furo",
   "sphinx-copybutton",
+  "six",
+  "typing_extensions",
+  "charset_normalizer",
```
lonbar: I don't think we should include this.

Which sphinx extensions require this? If these packages are required by a sphinx extension they should be installed automatically when the extension is installed. If the dependencies change, there is no point to install these packages.

.github/workflows/ci.yml
nikosarcevic: docs updated in 6ddb89f4585ee9178f4ed6de36e7838aa069edc2

pyproject.toml
nikosarcevic: reverted in 135cef2a604cec8b6bc8afb186c0262b646a7467

lonbar (approved):

PR #473, opened 2026-09-13
Add symmetrisation option for DALI tensors

lonbar:
Adds the option `symmetrize_dali` to `forecasting.forecast_core:get_forecast_tensors()`. It symmetrises the DALI tensors across all axes. The DALI tensors are by construction symmetric with respect to permuting any pair on indices, and this option makes this explicit.

The option can be adjusted by passing `symmetrize=True` to `ForecastKit.dali()`. The default has been set to `True`, which results in different behaviour from what is available in current releases of Derivkit.


src/derivkit/forecasting/forecast_core.py
```diff
@@ -161,14 +164,19 @@ def get_forecast_tensors(
...
-                np.einsum(
+            dali_tensor = np.einsum(
                     contractions[order1][order2],
                     derivatives[order1],
                     invcov,
                     derivatives[order2],
-                ).astype(np.float64, copy=False)
-            )
+            ).astype(np.float64, copy=False)
+            if symmetrize_dali:
```
nikosarcevic, line 174: am a bit worried about the factorial scaling here. For the rank-6  T_{(3,3) tensor this loops over 720 permutations of the full array, and sum(...) also repeatedly allocates intermediate arrays. can  we prehaps move this into a dedicated helper and at least accumulate into a preallocated array? We may also be able to exploit the intrinsic derivative/block symmetries to reduce the number of unique permutations substantially.

src/derivkit/forecasting/forecast_core.py
```diff
@@ -69,6 +71,7 @@ def get_forecast_tensors(
         method: Method name or alias (e.g., ``"adaptive"``, ``"finite"``).
             If ``None``, the :class:`derivkit.derivative_kit.DerivativeKit`
             default (``"adaptive"``) is used.
+        symmetrize_dali: Flag to force symmetrization across all DALI tensor axes.
```
nikosarcevic: Could we make this a little more explicit about the semantics? In particular, True returns the fully symmetrized DALI tensors, while False returns the raw derivative contractions before symmetrization.
I’d suggest wording like:
```
symmetrize_dali:
    If True, fully symmetrize each DALI tensor over its parameter
    axes. If False, return the raw derivative contractions.
```

src/derivkit/forecasting/forecast_core.py
nikosarcevic: minor thing: do we want this to apply to Fisher as well? J.T @ C^-1 @ J should already be symmetric by definition, so perhaps this should only be applied for order1 > 1. That would also keep the Fisher construction exactly unchanged.

tests/test_forecast_forecast_core.py
```diff
@@ -589,6 +600,162 @@ def model(th):
...
+            raise ValueError("Untested values added to SUPPORTED_FORECAST_ORDERS")
+
+    # The _get_derivatives() method in get_forecast_tensors() is replaced by dummy(),
+    # so we can purely test the symmetrisation.
+    # The function passed to get_forecast_tensors() is a dummy function which is not
+    # used to determine the output. The covariance matrix represents a 1D model.
+    function_to_patch = "derivkit.forecasting.forecast_core._get_derivatives"
+    # _get_derivatives() is not called directly so we need to ignore the linter warning.
+    with patch(function_to_patch, wraps=dummy) as mock_get_derivatives: #noqa
+        result = get_forecast_tensors(
```
nikosarcevic, line 743: snicethis test specifically validates the new symmetrization behavior, can we pass symmetrize_dali=True explicitly here rather than relying on the default? That makes the purpose of the test clear and keeps it independent of any future change to the default.

tests/test_forecast_forecast_core.py
```diff
@@ -589,6 +600,162 @@ def model(th):
...
 
 
+def test_dali_symmetries():
+    """Tests that the DALI tensors are invariant under any permutation of the axes.
+
+    The difference with test_expected_symmetries() is that this test checks that
+    the tensors are invariant under any permutation of the axes. No derivatives are
+    calculated.
+    """
+    # The "derivative tensors", mimicking a 1D model containing 2 variables.
```
nikosarcevic: ```# Mock derivative tensors for a model with one observable and two parameters.```

tests/test_forecast_kit_dali.py
```diff
@@ -431,12 +431,15 @@ def test_build_dali_matches_analytic_reference_across_methods(
 
     g_expected, h_expected = analytic_reference_dali(theta0, cov)
 
+    # The symmetrization is turned off here, as this is tested in
+    # tests.test_forecase_forecast_core.py:test_dali_symmetries()
```
nikosarcevic: Tiny typo in the test reference: should this be test_forecast_forecast_core.py

tests/test_forecast_kit_dali.py
nikosarcevic, line 43: Since build_dali now exposes symmetrize_dali, could we also record/assert that this argument is forwarded correctly here? At the moment it is absorbed into **dk_kwargs, so a regression where build_dali stops forwarding it would not be caught explicitly.

src/derivkit/forecasting/forecast_core.py
lonbar: I think that this is premature optimisation. Consider the following comparison:
```
(venv) $ cat sym_speed_test.py 
from argparse import ArgumentParser

from derivkit.forecast_kit import ForecastKit
import numpy as np

def parse_arguments():
    parser = ArgumentParser()
    parser.add_argument(
        "--symmetrize",
        action="store_true",
        default=False
    )
    return parser.parse_args()

def f(x):
    return np.array([
        np.cos(np.exp(x[0] - x[1])),
        -1/np.cosh(np.log(x[2])),
        np.prod(x)
    ])

def main(symmetrize):
    x0 = np.array([1.2, 0.3, 0.49])
    fk = ForecastKit(f, x0, np.eye(3))
    dali = fk.dali(
        symmetrize=symmetrize,
        forecast_order=3
    )


if __name__ == "__main__":
    args = parse_arguments()
    main(args.symmetrize)
(venv) $ time python sym_speed_test.py --symmetrize

real	0m4.001s
user	0m5.404s
sys	0m0.540s
(venv) $ time python sym_speed_test.py

real	0m4.059s
user	0m5.449s
sys	0m0.560s
```
There is no actual difference (the actual time is subject to some run-time variation, but the point still stands).

While the `symmetrize_dali` option isn't necessary for Fisher I would not make an exception for it, as there is no substantial benefit for doing this.

tests/test_forecast_forecast_core.py
lonbar: Yes, good point!

lonbar: Updated in https://github.com/derivkit/derivkit/pull/473/commits/a576f904c16bfaea899f91d828279d56ab5f02f1.

src/derivkit/forecasting/forecast_core.py
lonbar: I wonder if this raises more questions than it answers. Which derivative contractions? Maybe "force" can be replaced with "guarantee"?

tests/test_forecast_kit_dali.py
lonbar: Corrected in https://github.com/derivkit/derivkit/pull/473/commits/be04bc8926d7aed68ef4714585a015b0b82fdc02.

tests/test_forecast_forecast_core.py
lonbar: Added in https://github.com/derivkit/derivkit/pull/473/commits/ce2811df84f31fbb62e4c439f7fc1336fe2a8fd1.

tests/test_forecast_kit_dali.py
lonbar: Yes, good catch. Added in https://github.com/derivkit/derivkit/pull/473/commits/124b523ff1d890feb0127861a0a72f7e8adeb508.

src/derivkit/forecasting/forecast_core.py
nikosarcevic: I mean it is much different now in terms of waht it returns compared to just dping contraction of individual terms (essentially a combo of derivs is what I mean). i think being more explicit here is better

src/derivkit/forecasting/forecast_core.py
nikosarcevic: ok if you feel strongly we can keep as is, it is fine.

src/derivkit/forecasting/forecast_core.py
lonbar: Changed in https://github.com/derivkit/derivkit/pull/473/commits/9b7e8ad153536a8692b4fcf8e032c845b728a647.

nikosarcevic (approved):
Thank you for doing this, much appreciated and useful.

PR #474, opened 2026-09-13
Change derivative backend

lonbar:
Updates the default derivative backend from the adaptive polynomial to local polynomial method.


src/derivkit/forecasting/forecast_core.py
```diff
@@ -302,22 +302,15 @@ def _vectorize_model_output(theta: ArrayLike1D) -> NDArray[np.float64]:
 
     ckit = CalculusKit(_vectorize_model_output, theta0_arr)
 
-    if order >= 2:
-        derivkit_logger.info(
-            "[DALI BACKEND POLICY] order=%s: deliberately using finite differences "
-            "with Ridders extrapolation for higher-order derivatives because "
-            "polyfit/adaptive are too effective at smoothing and can wash out "
-            "the local higher-order curvature that DALI is meant to capture.",
+    if order >= 2 and method != "finite":
```
nikosarcevic, line 305: can we add a small unit test for this? just a smokiest, something simple?

lonbar: After discussing offline we decided not to do this.

nikosarcevic (approved):

PR #476, opened 2026-09-14
Patch chi2

### LINKED ISSUE #475: Fix DALI doublet coefficients in build_delta_chi2_dali
the current DALI posterior evaluator appears to use the wrong coefficients for the doublet terms.

For the derivkit convention

$$
\log p = -\frac{1}{2}\Delta\chi^2,
$$

the doublet contribution should be

$$
\Delta\chi^2 =F_{ab}\delta^a\delta^b+ D^1_{abc}\delta^a\delta^b\delta^c+ \frac{1}{4}D^2_{abcd}\delta^a\delta^b\delta^c\delta^d.
$$

The current implementation uses

```python
(1 / 3) * D1[d^3]
(1 / 12) * D2[d^4]
```

so both doublet contributions are smaller by a factor of 3.

They should be

```python
D1[d^3]
(1 / 4) * D2[d^4]
```

The triplet coefficients appear to be correct.

Before changing the evaluator, we should confirm that `get_forecast_tensors` does not already absorb these multiplicity factors into `D1` and `D2`.

[see here](https://github.com/derivkit/derivkit/blob/f4241d336eec1ba4096fd73f3bbf21ba54c997e9/src/derivkit/forecasting/expansions.py#L400)

lonbar:
Updates the numerical coefficients in the chi-squared polynomial.

Solves #475.


nikosarcevic (approved):
thanks!

PR #479, opened 2026-09-14
Display correct coefficients in chi-squared docstring

### LINKED ISSUE #477: fix module string in forecasting expansions
in #476 we fixed the polynomial coeeficients but forgot to [update the module string](https://github.com/derivkit/derivkit/blob/8bd800e07dd1d2ff90250a73d006e2ec0689aabb/src/derivkit/forecasting/expansions.py#L26)

should be a straightforward fix.

lonbar:
With this patch the changes in Chi-squared distribution introduced in 8bd800e will be shown in `derivkit.forecasting.expansions.build_delta_chi2_dali()`. These had not been propagated to the module's docstring in the patch. As a result, a user would get a different answer from what is expected.

Solves #477.


nikosarcevic (approved):
TYSM!!!

PR #483, opened 2026-09-19
Generalize hyper-Hessian to higher derivative orders

nikosarcevic:
## Summary

Generalizes the hyper-Hessian implementation to support arbitrary positive
derivative orders while preserving third order as the default.

## Changes

- add an `order` argument to `build_hyper_hessian`
- generalize unique derivative index construction and tensor symmetrization
- generalize mixed partial derivative evaluation
- add tests for first, second, third, and fourth derivative orders
- preserve the existing third order behavior by default

## Testing

Full test suite passes.


src/derivkit/calculus/hyper_hessian.py
```diff
@@ -66,6 +67,9 @@ def build_hyper_hessian(
     if theta.size == 0:
         raise ValueError("theta0 must be a non-empty 1D array.")
 
+    if order < 1:
+        raise ValueError("Derivative order must be at least 1.")
```
lonbar: It might be worth allowing the value 0, which returns the values of the original function.

lonbar (commented):
This would be a nice new feature to have. I think that what you did are sensible generalisations of what the hyper hessian does now, so it looks good to me. I just have one comment, which is not really important.

I added https://github.com/derivkit/derivkit/pull/483/commits/42cefc15b82edabe3a9ca0ae4258bcea1496ff42, to test the hyperhessian for a function that is a bit more complicated than a scalar-valued function.

One thing to keep in mind is that now we don't test hyperhessians of order higher than 4. It's not entirely clear to me how to sensibly do this, or what orders we want to support.

tests/test_calculus_hyper_hessian_cache.py
```diff
@@ -1,55 +1,82 @@
...
-def test_hyper_hessian_cache_does_not_change_result():
-    """Tests that hyper-Hessian caching does not change the result."""
+_METHOD_CASES = [
+    ("finite", {}),
+    ("finite", {"extrapolation": "richardson"}),
+    ("finite", {"extrapolation": "ridders"}),
+    ("finite", {"extrapolation": "gauss-richardson"}),
+    ("adaptive", {}),
+    ("local_polynomial", {}),
+]
```
lonbar, line 15: We might want to import this from some test utility, as this object is used in other modules too.

tests/test_calculus_hyper_hessian.py
```diff
@@ -18,13 +18,10 @@
...
-    ("finite", {"extrapolation": "richardson"}, 2e-3, 1e-4),
-    ("finite", {"extrapolation": "ridders"}, 2e-3, 1e-4),
-    ("finite", {"extrapolation": "gauss-richardson"}, 1e-3, 1e-4),
-    ("adaptive", {}, 1e-4, 1e-3),
-    ("local_polynomial", {}, 1e-4, 1e-4),
+_SMOOTH_METHOD_CASES = [
+    ("finite", {}),
+    ("adaptive", {}),
+    ("local_polynomial", {}),
 ]
```
lonbar: This removes a lot of test cases; what's the reason for that?

src/derivkit/calculus/hyper_hessian.py
nikosarcevic: ah good one. added here 85b9d8e075a1986b9d6e9fff02d0a49204e5b30d

nikosarcevic: also added a test for it forgot to say

tests/test_calculus_hyper_hessian_cache.py
nikosarcevic: we can raise this as an issue but not fix in this pr as these show up a lot so this pr would be huge. WDYS?

tests/test_calculus_hyper_hessian.py
nikosarcevic: this was put back in.

tests/test_calculus_hyper_hessian_cache.py
lonbar: Set up issue #485.

lonbar (approved):
LGTM!

